PR #148, opened 2012-03-18
refactored to use a storage backend

chrisle:
All I did here was just move VCR.Configuration.cassette_library_dir to VCR.Cassette.StorageBackends[:file_system]

What I'm experimenting with my fork is something like this:

```
# Remain backward compatible with older configs
# Assumes c.storage_backends = :file_system
VCR.configure do |c|
  c.cassette_library_dir = 'spec/cassettes'
end

# Store to a DB
VCR.configure do |c|
  c.storage_backends = :db
  # ???
end

# Store to BOTH disk and DB
VCR.configure do |c|
  c.storage_backends = :file_system, :db
  c.cassette_library_dir = 'spec/cassettes'
end

# ... meanwhile in someone's application ...
require 'vcr'
require 'mysql2'

module MyApp
  ....
end

module VCR
  class Cassette
    class StorageBackends
      module Db
        extend self

        def storage_location
          # return a string? the database name?
        end

        def storage_location=(db)
          # connect to your db?
        end

        def [](key)
          @mydb.find_by_key(key)
          # etc..etc.
        end

        def []=(key, value)
          @mydb.update_attribute(key: value)
          # etc...etc.
        end

      end
    end
  end
end

```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 [Full Changelog](http://github.com/myronmarston/vcr/compare/v2.0.0...master)
 
+* Add new `:use_scenario_name` option to the cucumber tags API. This
+  allows you to use a generic tag (such as `@vcr`) and have the
+  cassettes named based on the feature and scenario rather than based on
+  the tag. Thanks to [Omer Rauchwerger](https://github.com/rauchy) for
+  the implementation and [Chad Jolly](https://github.com/cjolly) for the
+  initial idea and feedback.
+* Add new `:decode_compressed_response` cassette option. When set to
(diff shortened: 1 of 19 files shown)


lfv89:
Correct me if I'm wrong but VCR is still missing this feature correct? @chrisle @myronmarston

myronmarston:
> Correct me if I'm wrong but VCR is still missing this feature correct? @chrisle @myronmarston

I've not worked on VCR in like 5 years or so but it looks like the feature mentioned above is still there:

https://github.com/vcr/vcr/blob/v6.1.0/lib/vcr/cassette/persisters/file_system.rb

If you have more questions, please open a new issue and direct it to the current maintainers.

Issue #378, opened 2014-01-22
Rspec metadata integration names file '.yml' when using rspec one-liner syntax


jspooner:
When using the [Rspec metadata integration](https://www.relishapp.com/vcr/vcr/v/2-4-0/docs/test-frameworks/usage-with-rspec-metadata) and one-line syntax `it { should be_valid }` VCR names the cassette file ".yml".  This was quite confusing at first file is invisibe yet vcr still works.

Rspec is able to take a test like `it { longitude.should == -122.0839544 }` and generate a description like `Model initialization with longitude should == -122.0839544`.  I believe the file name should be something like 'should_eql_122.0839544.yml'.

Here is how I see this fitting into [rspec_metadata.feature](https://github.com/vcr/vcr/blob/master/features/test_frameworks/rspec_metadata.feature).

``` diff
diff --git a/features/test_frameworks/rspec_metadata.feature b/features/test_frameworks/rspec_metadata.feature
index b0d296c..fc64d1b 100644
--- a/features/test_frameworks/rspec_metadata.feature
+++ b/features/test_frameworks/rspec_metadata.feature
@@ -76,6 +76,8 @@ Feature: Usage with RSpec metadata
           make_http_request.should == 'Hello'
         end

+        it { make_http_request.should == 'Hello' }
+
         context 'in a nested example group' do
           it 'records another one' do
             make_http_request.should == 'Hello'
@@ -93,6 +95,7 @@ Feature: Usage with RSpec metadata
     Then it should pass with "4 examples, 0 failures"
      And the file "spec/cassettes/VCR_example_group_metadata/records_an_http_request.yml" should contain "Hello"
      And the file "spec/cassettes/VCR_example_group_metadata/records_another_http_request.yml" should contain "Hello"
+     And the file "spec/cassettes/VCR_example_group_metadata/should_be_Hello.yml" should contain "Hello"
      And the file "spec/cassettes/VCR_example_group_metadata/in_a_nested_example_group/records_another_one.yml" should contain "Hello"
      And the file "spec/cassettes/VCR_example_metadata/records_an_http_request.yml" should contain "Hello"
```

tessi:
I've opened a PR for another approach -- we could use rspecs nested scope for name the cassette -- see #514 for details.

PR #382, opened 2014-01-27
Mark vcr test that use one line rspec syntax as pending

jspooner:
I have two solutions for https://github.com/vcr/vcr/issues/378

I prefer this option that marks the test as pending rather then raising an error.


krainboltgreene:
Going to close this for now, since it's so far behind. Just buzz me and I'll reopen it if you want!

Issue #428, opened 2014-07-25
TypeError: no implicit conversion of Symbol into String


mrinterweb:
I keep running into errors on line 125 of lib/vcr/structs.rb
TypeError: no implicit conversion of Symbol into String

`new_headers[String.new(k)] = convert_to_raw_strings(val_array)`

The issue is when `k` is a Symbol. I found the following change fixes my issue:

`new_headers[k.to_s] = convert_to_raw_strings(val_array)`

The problem was when I was using the Excon gem as so

```
@socket_server_connection ||= Excon.new socket_server_url,
  headers: { API_KEY: 'myCoolAPIKeyGoesHere' },
  instrumentor: ActiveSupport::Notifications,
  instrumentor_name: 'socket_server_connection'
```

The `:API_KEY` symbol in the headers was messing it up. I'll submit a pull request with the fix.

PR #429, opened 2014-07-25
VCR::Normalizers::Header#normalize_headers now can accept symbols as key...

mrinterweb:
This fixes issue #428

### DIFF
--- a/lib/vcr/structs.rb
+++ b/lib/vcr/structs.rb
@@ -119,7 +119,7 @@ module VCR
             else [v]
           end
 
-          new_headers[String.new(k)] = convert_to_raw_strings(val_array)
+          new_headers[k.to_s] = convert_to_raw_strings(val_array)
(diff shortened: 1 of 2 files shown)


Ch4s3:
So I think this should be reopened. I totally agree that header keys should generally be strings, there are a handful of gems out there that use `symbols`. See for example the default example from [Typhoeus](https://github.com/typhoeus/typhoeus):
````ruby
request = Typhoeus::Request.new(
  "www.example.com",
  method: :post,
  body: "this is a request body",
  params: { field1: "a field" },
  headers: { Accept: "text/html" }
)
````

This breaks VCR, obviously. I think its quite easy in a number of http libraries to, in ignorance, use `key: value` headers and they pass that along just fine. Sure, that isn't VCR's fault, but I think some level of input sanitization is a nice thing to do.

If this truly can't be merged in, could we get a warning in the readme or docs?

PR #457, opened 2014-12-06
Updated aruba_timeout_seconds to 60 seconds.

andhapp:
An attempt to fix the issue on travis by bumping the `timeout` for child process.


krainboltgreene:
If you merge from core we can retry this.

PR #459, opened 2014-12-14
Opt-in to new travis environment.

myronmarston:
This is a follow up to #458.


mcfiredrill:
Is `sudo: false` still needed?

mislav:
@myronmarston On the new infrastructure you might be able to use built-in bundle caching features and get rid of my `cached-bundle` script.

@mcfiredrill I think it is. It might be a default only for newer repos.

mislav:
Done in 5a3efb80e4be13badb8411261c9d6e45948a28a1

PR #470, opened 2015-01-12
Allow matchers in VCR cassettes to be regular expressions.

mikejihbe:
This was documented, but didn't actually work in practice. This change allowed us to use the feature as planned.

### DIFF
--- a/lib/vcr/request_matcher_registry.rb
+++ b/lib/vcr/request_matcher_registry.rb
@@ -109,10 +109,10 @@ module VCR
     end
 
     def register_built_ins
-      register(:method)  { |r1, r2| r1.method == r2.method }
-      register(:uri)     { |r1, r2| r1.uri == r2.uri }
-      register(:body)    { |r1, r2| r1.body == r2.body }
-      register(:headers) { |r1, r2| r1.headers == r2.headers }
+      register(:method)  { |r1, r2| r2.method.is_a?(Regexp)  ? r1.method.match(r2.method)   : r1.method == r2.method }
+      register(:uri)     { |r1, r2| r2.uri.is_a?(Regexp)     ? r1.uri.match(r2.uri)         : r1.uri == r2.uri }
+      register(:body)    { |r1, r2| r2.body.is_a?(Regexp)    ? r1.body.match(r2.body)       : r1.body == r2.body }
+      register(:headers) { |r1, r2| r2.headers.is_a?(Regexp) ? r1.headers.match(r2.headers) : r1.headers == r2.headers }
 
       register(:host) do |r1, r2|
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Is it me or does this scream for a separate method that takes something (like `r2.{method,uri,body,headers}`) and performs this check so all you do is 

```rb
[:method, :uri, :body, :headers].each do |attr|
  register(attr) { |r1, r2| check(r1.public_send(attr), r2.public_send(attr)) }
end
```

krainboltgreene:
Agreed.

myronmarston:
FWIW, VCR 1.x supported regexes out-of-the box, but that functionality was purposefully removed in VCR 2.0 in favor of the much more flexible custom request matchers. @mikejihbe you mention that this is documented to work but I'm not aware of any VCR 2.0 docs that state that regexes work.  Can you point us to where you saw that?

Anyhow, as you've demonstrated here, any project that wants to use regexes can get that feature simply by registering a custom request matcher that uses regex matching.  In general, I think regexes are inferior to custom request matchers because they are less flexible than matchers (which are essentially lambdas, and can have any arbitrary logic) and require that you hand-edit a VCR cassette, which inhibits future re-recording.  Thus, I specifically wanted to discourage regex use in VCR 2.0 and chose to remove the feature.

@krainboltgreene is maintaining VCR now and he may choose to add back support for regexes (which would be fine) but I wanted to give the background about why this was functionality was removed in VCR 2.0.

krainboltgreene:
Excuse my ignorance, but does anyone have an example of this in use?

myronmarston:
> Excuse my ignorance, but does anyone have an example of this in use?

I'm not quite sure what your "this" is referring to, but assuming you mean custom request matchers....yes.  They are frequently used in the wild:

https://github.com/search?q=register_request_matcher&type=Code&utf8=%E2%9C%93

The most common use of custom request matchers is to ignore non-deterministic query params that are not part of the identity of the request; the `uri_without_params` method is widely used for that:

https://github.com/search?utf8=%E2%9C%93&q=uri_without_param&type=Code&ref=searchresults

The VCR wiki has a page with a listing of common request matchers:

https://github.com/vcr/vcr/wiki/Common-Custom-Matchers

And, FWIW, the background on this feature is on my blog: http://myronmars.to/n/dev-blog/2011/10/custom-request-matchers-in-vcr-2-0

mcfiredrill:
I use a lot of custom request matchers. 

I'm just wondering if @mikejihbe uses them, and if so are supporting regexes a better solution for you in some way?

mcfiredrill:
Still pretty 👎  on this feature in favor of custom request matchers. Please reopen if you disagree.

PR #485, opened 2015-04-13
Fix interaction mismatch bug

calebwoods:
This change fixes an issue in an app I'm working on where, with a lot of requests coming into VCR at the same time, using `delete_at` in the `HTTPInteractionList` class seems to cause a mismatch in interactions and VCR returns the wrong stubbed interaction or `nil`.

### Debugging

Here is some code I added to the `HTTPInteractionList` class to debug the issue.

```ruby
def response_for(request)
 if index = matching_interaction_index_for(request)
   interaction = @interactions.delete_at(index)
   @used_interactions.unshift interaction

   # start debugging
   if request != interaction.request || interaction.nil?
     puts "Request: #{request}"
     puts "Interaction: #{interaction.request}"
   end
   # end debugging

   log "Found matching interaction for #{request_summary(request)} at index #{index}: #{response_summary(interaction.response)}", 1
   interaction.response
 elsif interaction = matching_used_interaction_for(request)
   interaction.response
 else
   @parent_list.response_for(request)
 end
end
```

With that change I can see this type of output in my Spinach tests using Poltergeist.

```
Scenario: Hidden Almac settings
    ✔  Given login to settings as 'dispense_user'  # features/steps/common_steps/login.rb:18
Request: #<struct VCR::Request method=:get, uri="http://localhost:8080/ws/projects/285/menus/designer", body="", headers={"authorization"=>["Basic WDozMzNkYmY4Yy00YTNjLTRiZjItOGNhMi0xNTM1Y2ZkNDA2YTI="]}>
Interaction: #<struct VCR::Request method=:get, uri="http://localhost:4100/api/v1/studies/285", body="", headers={"authorization"=>["Basic MzMzZGJmOGMtNGEzYy00YmYyLThjYTItMTUzNWNmZDQwNmEyOlg="]}>
    !  And on the design settings page             # features/steps/design_settings.rb:5
    ~  When study is not using almac settings
    ~  Then can't see Almac settings

        One or more errors were raised in the Javascript code on the page. If you don't care about these errors, you can ignore them by setting js_errors: false in your Poltergeist configuration (see documentation for details).

        TypeError: null is not an object (evaluating 'rawMenuItems['menuItems'].filter')
        TypeError: null is not an object (evaluating 'rawMenuItems['menuItems'].filter')
            at http://127.0.0.1:31338/assets/application.js:78570
            at http://127.0.0.1:31338/assets/application.js:21016 in wrappedCallback
            at http://127.0.0.1:31338/assets/application.js:21102
            at http://127.0.0.1:31338/assets/application.js:22145 in $eval
            at http://127.0.0.1:31338/assets/application.js:21957 in $digest
            at http://127.0.0.1:31338/assets/application.js:22249 in $apply
            at http://127.0.0.1:31338/assets/application.js:17784 in done
            at http://127.0.0.1:31338/assets/application.js:17998 in completeRequest
            at http://127.0.0.1:31338/assets/application.js:17937 in onreadystatechange
        /Users/caleb/code/rolemodel/merge/dispense/vendor/bundle/ruby/2.1.0/gems/poltergeist-1.6.0/lib/capybara/poltergeist/browser.rb:323:in `command'

Error summary:
  Errors (1)
    Design Settings :: Hidden Almac settings :: And on the design settings page
```

As you can see the `request` coming into VCR and the request for the interaction do not match.  My guess is that this line of code is the issue.

```ruby
interaction = @interactions.delete_at(index)
```

I suspect that somehow between when the index is found in the line above and when the `delete_at` runs that the array has been changed and the index now points to a new object.  Unfortunately I've not been able to test completely deterministically as only fails when I run my entire suite and on different tests each time.

### Solution

The code in this PR is an attempted solution, which seems to be working on my application, but I'm not sure the best way to test or if there is a simpler approach.  I've checked that the tests pass, other than the 2 that are already failing on master. Feedback welcome.


krainboltgreene:
Wow, impressive!

myronmarston:
FWIW, VCR is not threadsafe.  It sounds like the only way you could hit this is if you were using VCR from multiple threads.  It would be nice to get VCR threadsafe but I think that should be addressed wholistically and not in a one-off manner like this.  You're going to run into other problems if you use VCR from multiple threads.

krainboltgreene:
I have to agree with @myronmarston. I'm tempted to take this pull request simply because my goal for 3.0.0 is thread safety and this is a stop gap measure...But since I'm not able to currently work on anything heavy due to computer issues I can't know when that'll happen.

Let me think about it.

calebwoods:
Makes sense that this issue would be related to threadsafety.  Is there already some current thinking about what would need to change make VCR threadsafe?  Or how one would write some failing tests for the issue?

My team is currently using VCR on several projects with Poltergeist/PhantomJS, which I think means we could potentially run into this a lot.  Given that I'd be very interested in trying to help move the ball forward on a proper solution.

krainboltgreene:
No, I haven't been able to see what would need to change in order to make VCR threadsafe. I'm open to help, though.

krainboltgreene:
@calebwoods Hey, any chance you can update this PR? We had a major patch that might have fixed this.

calebwoods:
@krainboltgreene using the latest commit on master (cc57fc4c6da01c37ecb5700040a5eeda5633e12a) it looks like that's mostly resolved the issue I saw on 2.9.3.

I'm seeing some other failures, but not entirely sure if those are the test suite issues or something new that was introduced.  I'll try to investigate further and report back.

mcfiredrill:
@calebwoods Have the recent thread safety patches solved your issues? If not, can you open a new issue if there are new problems?

PR #493, opened 2015-06-06
Disable VCR by default

dcrec1:
In a TTDed application, a Rails one for example, tests may be correctly passing but when deployed to production, the app may break because the VCR gem was loaded for all environments and is turned on by default. This can easily happen if somebody forgets to add `group: :test` when specifying the gem in the Gemfile.

This pull request turns VCR off by default. The bad thing of this change is that one more step will be needed in the setup, but the good thing is that it will be not possible to break an application just by adding the gem.

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,7 @@ require 'rubygems'
 require 'test/unit'
 require 'vcr'
 
+VCR.turn_on!
 VCR.configure do |config|
   config.cassette_library_dir = "fixtures/vcr_cassettes"
   config.hook_into :webmock # or :fakeweb
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -340,7 +340,7 @@ private
   end
 
   def initialize_ivars
-    @turned_off = false
+    @turned_off = true
   end
 
   initialize_ivars # to avoid warnings
--- a/spec/lib/vcr_spec.rb
+++ b/spec/lib/vcr_spec.rb
@@ -341,9 +341,9 @@ describe VCR do
   end
 
   describe '.turned_on?' do
-    it 'is on by default' do
+    it 'is off by default' do
       VCR.send(:initialize_ivars) # clear internal state
-      expect(VCR).to be_turned_on
+      expect(VCR).to_not be_turned_on
     end
   end
 end


sigmavirus24:
I'm strongly :-1:  on this. It isn't VCR's job to prevent a developer from deploying it into production accidentally.

dcrec1:
Ok, maybe it's not VCR's current job to do certain task, but why is this a problem? I mean, will this change make VCR harder to maintain, will make VCR harder to use? What's the negative effect of this change? Do you think there is no gain? I think this change will make VCR a little bit harder to use, but also will prevent VCR from breaking applications just by adding it. I don't think the current behaviour is wrong, I just like code that is hard to break. If you think there is some gain, but the loss is bigger, what do you think of automatically enabling VCR when called `VCR.configure` without having previously explicitly turned VCR off?

myronmarston:
It's a problem because it's a breaking change for every VCR user out there.  Many tutorials (and published books, such as [Rails 4 Test Prescriptions](https://pragprog.com/book/nrtest2/rails-4-test-prescriptions) and [Cucumber Recipes](https://pragprog.com/book/dhwcr/cucumber-recipes)) exist which show new users how to setup VCR and `VCR.turn_on` isn't one of the steps they tell new users to do.  This change means all the existing tutorials no longer work for new users.

More than that, ruby already provides a system whereby gems are disabled by default and you can opt-in to enable them--`require`.  I don't see a reason for VCR to provide an additional system for that: now users have to enable VCR with two systems (`require` and `VCR.turn_on`) for it to work.  In general, I think that `Bundler.require` is an anti-pattern, [as I've blogged about](http://myronmars.to/n/dev-blog/2012/12/5-reasons-to-avoid-bundler-require), and this is yet another reason to avoid it.  I don't think we should make a breaking change to VCR to encourage something that I consider an anti-pattern.

donv:
Putting Vcr in the "test" group and having it work directly is a very important feature which should not be removed.  I also :-1: this.

dcrec1:
@myronmarston ok, I don't agree but I understand your motivations. I'm closing the pull request, thanks for taking a look at this :)

PR #495, opened 2015-06-20
rbenv complains if version specified is not exact

asymmetric:
There are several matches for 2.1: 2.1.0, 2.1.1, etc.

rbenv/ruby-install is more precise than other ruby version managers (ahem) and doesn't proceed with the installation.


tylercollier:
+1. Yes, it's curious as to why 2.1 was chosen. Why not e.g. 2.1.0?

sigmavirus24:
Likely because the developers don't use rbenv. I know that at least rvm will pick the latest 2.1.* if you only put 2.1 in the ruby-version file. This means that you can have the latest 2.x.* with all the latest CVE fixes without having to also carry old, out of date, and vulnerable versions of ruby.

krainboltgreene:
Grumble grumble. Yes I don't use rbenv, yes it's because I assumed it would handle like `~> 2.1.0`, since that's how Ruby's versioning works now.

krainboltgreene:
I don't want to alienate rbenv users, so I'm just going to suck it up and install a specific verison of ruby for vcr. Sorry for this problem!

mislav:
@krainboltgreene Could we just remove `.ruby-version`? This library is compatible with all currently maintained Ruby releases. It doesn't matter which one somebody uses in development.

krainboltgreene:
I mean, I like having `.ruby-version`, but yes, you're right in that any ruby version should currently work with vcr.

mislav:
Yeah, let's not force people to install a specific Ruby version when not necessary.

PR #497, opened 2015-08-04
add compatibility with Cucumber 2 for use_scenario_name option

harryw:
I've found that the VCR integration with Cucumber doesn't play nicely with Cucumber 2, and it seems to be down to a change in the way Cucumber `feature.name` works in Cucumber 2.

I've looked for a changelog entry in Cucumber to explain this without luck, and it's arguable that this is a bug in Cucumber, but this seems like the most pragmatic solution in any case.


harryw:
The test failures are not new, FWIW.

PR #500, opened 2015-08-12
Dependency fix

chriswoodrich:
Turns out it was an issue with the new version of Faraday.  Downgrading to 0.9.0 fixed it.  Your CI may not be working but here's the results from my machine:

```2110 examples, 0 failures, 1 pending```

### CODE REVIEW

**Comment by mcfiredrill** [comment]
@chriswoodrich 
Thanks for taking a shot at this. It looks like introducing the Gemfile.lock caused the build to fail on some other versions of Ruby.

```
Gem::InstallError: cucumber-core requires Ruby version >= 1.9.3.
Using vcr 2.9.3 from source at .
An error occurred while installing cucumber-core (1.2.0), and Bundler cannot
continue.
Make sure that `gem install cucumber-core -v '1.2.0'` succeeds before bundling.
```

I'm not sure, but is it necessary to include the Gemfile.lock?

chriswoodrich:
@mcfiredrill I feel fairly strongly about keeping the Gemfile.lock, but after reading this article: http://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/  I think it makes sense to remove it, and manually specify working dependencies in the gemspec.  

I'll keep this PR open and make those changes.

mcfiredrill:
@krainboltgreene any thoughts on this?

Looks like we're still failing on 1.9.2. I'm not sure if it's possible to get it passing without a separate Gemfile for 1.9.2 via appraisal.

krainboltgreene:
I think travis has that functionality built in, I'll start working on it tonight.

mcfiredrill:
Yeah maybe do something like this?
https://github.com/thoughtbot/appraisal#travis-ci-integration

krainboltgreene:
I'm currently moving to LA, so this is going slowly.

Issue #200, opened 2012-10-15
VCR is not threadsafe


Fryguy:
I've been having intermittent issues where VCR causes my tests to fail, raising UnhandledHTTPRequestError or causing deadlocks.

The code in question collects inventory from a third-party REST API, using rest-client.  Since we have to make about 30 API calls, we found that making each call in a separate thread gave us significant performance increases (~85% faster than unthreaded).

I recently started using VCR to add tests for this, and it seems the multi-threading is causing these errors.  Other tests that aren't multi-threaded do not exhibit these errors.

To narrow down and to prove it was the multi-threading, I added <code>allow_http_connections_when_no_cassette = false</code> and <code>allow_unused_http_interactions => false</code>, and was able to consistently get our test to fail with a different error.  Now, I get UnusedHTTPInteractionError, with a seemingly random set of unused interactions that change every time I run the spec.

The errors got me thinking that the bookkeeping around the VCR::Cassette::HTTPInteractionList may be involved.  For example, I added some puts instrumentation in VCR::Cassette#http_interactions, and I was able to show that on occasion, the right-side of the ||= statement was executed more than once.  I can also see that various methods in VCR::Cassette::HTTPInteractionList (like #response_for) do manipulations of that list.

I'm sure there are a bunch of places where synchronization needs to be added, but I haven't been able to narrow them all down.  The above are just some places I think might need it, but I'm not sure how to create a test for it.  I figured I'd open an issue to start discussing how to deal with this and to figure out how to make tests, as I am not sure even where to start.

carhartl:
FWIW, - for the particular case of having vcr hook into webmock - I achieved vcr threadsafety for the case of making several http requests in threads within an application.

After synchronizing access to a cassette's http interactions the remaining pain point was the `VCR::LibraryHooks::WebMock.with_global_hook_disabled` method, e.g. the `global_hook_disabled` state on the module. Somehow synchronizing access to it didn't work, and so I temporarily changed `externally_stubbed?` to simply return false. For the time being that removed the ability to stub requests with webmock next to using a cassette (we don't use that feature at the moment).

Is it worth it to look further into it?

With the fork our tests now run reliably again, and no bogus recordings are being made.

fotinakis:
FWIW, I was able to fix this for myself by using a mutex and synchronizing all HTTP threads handled by VCR:

```Ruby
require 'thread'

global_vcr_semaphore = Mutex.new

VCR.configure do |c|
  c.around_http_request do |request|
[...]

Issue #494, opened 2015-06-11
Turning off VCR inside around_http_request seems to cause strange fibre errors


brightbits:
We are turning off VCR based on the uri, like so:

```ruby
VCR.configure do |c|
  c.around_http_request(lambda {|r| r.uri =~ /<regexp>/}) do |request|
    VCR.turned_off(&request)
  end

  c.ignore_localhost = true
  c.cassette_library_dir = 'spec/support/fixtures/cassettes'
  c.hook_into :webmock
  c.configure_rspec_metadata!
  c.default_cassette_options = { record: :none }
end
```

Sometimes we get this error:
```
  undefined method `resume' for nil:NilClass
  
  /<gem path>/vcr-2.9.3/lib/vcr/configuration.rb:510:in `resume_fiber'
  /<gem path>/vcr-2.9.3/lib/vcr/configuration.rb:405:in `block in around_http_request'
  /<gem path>/vcr-2.9.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call'
  /<gem path>/vcr-2.9.3/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
  /<gem path>/vcr-2.9.3/lib/vcr/util/hooks.rb:14:in `conditionally_invoke'
  /<gem path>/vcr-2.9.3/lib/vcr/util/hooks.rb:29:in `block in invoke_hook'
  /<gem path>/vcr-2.9.3/lib/vcr/util/hooks.rb:28:in `map'
  /<gem path>/vcr-2.9.3/lib/vcr/util/hooks.rb:28:in `invoke_hook'
  /<gem path>/vcr-2.9.3/lib/vcr/library_hooks/webmock.rb:141:in `block in <module:WebMock>'
  /<gem path>/webmock-1.18.0/lib/webmock/callback_registry.rb:20:in `call'
  /<gem path>/webmock-1.18.0/lib/webmock/callback_registry.rb:20:in `block in invoke_callbacks'
  /<gem path>/webmock-1.18.0/lib/webmock/callback_registry.rb:15:in `each'
  /<gem path>/webmock-1.18.0/lib/webmock/callback_registry.rb:15:in `invoke_callbacks'
  /<gem path>/webmock-1.18.0/lib/webmock/http_lib_adapters/net_http.rb:89:in `block in request'
  /<gem path>/webmock-1.18.0/lib/webmock/http_lib_adapters/net_http.rb:98:in `call'
[...]

Issue #376, opened 2014-01-15
Intermittent exceptions caused by "undefined method `resume'"


bilts:
Running rspec tests using capybara against a Rails project produces intermittent exceptions coming from VCR's around_http_request implementation.  The error may appear on any tests and varies from run to run.

VCR Configuration in config/environments/test.rb

```
VCR.configure do |c|
  c.cassette_library_dir = 'fixtures/cassettes'
  c.hook_into :faraday
  c.around_http_request do |request|
    VCR.use_cassette('global', :record => :new_episodes, &request)
  end
end
```

Stack trace:

```
 Failure/Error: Unable to find matching line from backtrace
 NoMethodError:
   undefined method `resume' for nil:NilClass
 # /gem-root/gems/vcr-2.6.0/lib/vcr/configuration.rb:510:in `resume_fiber'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/configuration.rb:405:in `block in around_http_request'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/util/variable_args_block_caller.rb:9:in `call'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/util/variable_args_block_caller.rb:9:in `call_block'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/util/hooks.rb:14:in `conditionally_invoke'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/util/hooks.rb:29:in `block in invoke_hook'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/util/hooks.rb:28:in `map'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/util/hooks.rb:28:in `invoke_hook'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/request_handler.rb:51:in `invoke_after_request_hook'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/middleware/faraday.rb:108:in `invoke_after_request_hook'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/middleware/faraday.rb:48:in `ensure in handle'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/middleware/faraday.rb:48:in `handle'
 # /gem-root/gems/vcr-2.6.0/lib/vcr/middleware/faraday.rb:31:in `call'
 # /gem-root/gems/faraday_middleware-0.9.0/lib/faraday_middleware/response_middleware.rb:30:in `call'
[...]

Issue #368, opened 2013-12-12
Cassettes getting updated sporadically


swapab:
I observe this quite often, whenever the rspec run encounters this warning

```
WARNING: There appears to be a bug in WebMock's after_request hook
         and VCR is attempting to work around it. Some VCR features
         may not work properly.
```

then the cassette get updated for that respective test case.

VCR configuration looks like

``` ruby
VCR.configure do |c|
  c.ignore_hosts 'fake-example.com'
  c.configure_rspec_metadata!
  c.hook_into :webmock
  c.cassette_library_dir = 'spec/vcr'
  c.allow_http_connections_when_no_cassette = true
end
```

Am actually crawling a site, and recording the response

Example 1

``` ruby
   it "lets me crawl a site" do
      VCR.use_cassette "lets me import a site", record: :once do
        visit new_site_crawl_path
        within '.panel' do
          click_button 'crawl_site_btn'
        end
      end
    end
```

Then I get an updated Cassette even if the url is already visited.

Example 2

``` ruby
  context 'crawl pages', :vcr do
    before do
          login(@user)
          some.. more... stuff... 
          logout
          login_active_admin
    end

    it "limit the number of pages to be imported" do
      expect { click_button "Import Pages" }.to change(Page, :count)
    end
  end
```

Does the login-logout sequence can makes the difference ? like the request headers might be changing, just assuming. Though _Example 1_ doesn't have any such before block.

krainboltgreene:
Related to #200.

perlun:
I am seeing now also. Unfortunately also threaded code in my use case. Will try to disable threading for the specs, thanks for the suggestion.

ndbroadbent:
I just wanted to mention that I'm running into an issue where VCR is complaining with:

```
An HTTP request has been made that VCR does not know how to handle:
           - :record => :none
           - :match_requests_on => [:method, :uri]
...

Under the current configuration VCR can not find a suitable HTTP interaction
to replay and is prevented from recording new requests
```

I'm running an integration test with Capybara, Puma, and VCR. So it seems like this could be a thread-safety issue. The `method` and `uri` are identical so it should be finding the response in the cassette. The test is consistently failing on CI, but I haven't been able to reproduce it locally, even when running in an identical Docker image.

One thing I noticed is that all of the tests take about 2 minutes to run on CI, but it takes about 3 minutes on my local machine. So it seems like I'm only seeing this issue because the machine is too fast.

[...]

PR #512, opened 2015-10-10
Make VCR thread-safe

mrbrdo:
Trying to finally get rid of this problem. Seems to work well with my test case.
This is for the case when the code being tested is threaded, not for running tests in threads.

CC https://github.com/vcr/vcr/issues/200

### DIFF
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -22,6 +22,12 @@ module VCR
 
   extend self
 
+  # Mutex to synchronize access to cassettes in a threaded environment
+  CassetteMutex = Mutex.new
+
+  # The main thread in which VCR was loaded
+  MainThread = Thread.current
+
   autoload :CucumberTags,       'vcr/test_frameworks/cucumber'
   autoload :InternetConnection, 'vcr/util/internet_connection'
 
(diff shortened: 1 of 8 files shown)


myronmarston:
I just deactivated hound.  Sorry about the noise.

mrbrdo:
I wanted to get the specs passing for JRuby as well, but I can't figure out a way to only require curb, yajl-ruby and patron on MRI. I tried putting them into the Gemfile instead of the gemspec and that worked fine locally (and specs are all passing), but Travis then didn't use any of these gems even under MRI (bundle install did not include those gems in the output, and then the require failed in the specs).

mrbrdo:
I think this PR also solves #494 and #376

Fryguy:
:clap: Happy to see this!

mrbrdo:
Yes me too. :)

Thanks to Daniel Berger for the bounty which enabled this PR: http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-talk/433878

krainboltgreene:
Release out.

mrbrdo:
:fireworks:

perlun:
Sorry to be a party killer here, but... there are still issues. I'm running into #368 even with the latest version. Or is there anything I'm missing here, are there still threading issues when testing code that is multi-threaded?

PR #367, opened 2013-12-12
Re-enable rubinius on travis

JonRowe:
Similar to what we did on rspec/rspec-core ;)

Issue #213, opened 2012-11-25
Get green JRuby builds on Travis


myronmarston:
VCR used to get green builds on JRuby (and still works fine on JRuby) but it's been hard to keep the build consistently green on JRuby for both the 1.8 and 1.9 modes.  For now JRuby build failures are ignored.

I'd like to get it back to consistent green.

krainboltgreene:
The latest push opens the floodgates for failures.

mrbrdo:
It should work if you manage to make TravisCI only load curb, yajl-ruby and patron gems on MRI. I tried moving them to the Gemfile and using the `platforms` options, which worked locally but Travis did not install the gems (even on MRI).
The rspec specs should be passing otherwise after fixing Zlib as I have done in https://github.com/vcr/vcr/pull/512

mislav:
Please see https://github.com/vcr/vcr/pull/513 which avoids loading C-extension gems in Rubinius and JRuby

PR #513, opened 2015-10-15
Help tests pass in Rubinius and a step closer for JRuby

mislav:
These changes help tests pass for rbx and make the situation much better on JRuby, although there are still failures.

Fixes #367, references #213


mrbrdo:
I had no failures on JRuby 1.7.19 when I ran them on #512 (I did fix one minor issue first). Are they still failing for you even with #512 merged? I can look if I can help fix them if that's the case.

mislav:
@mrbrdo: You're right, it seems that [JRuby passes now](https://travis-ci.org/vcr/vcr/jobs/85552414) in CI for this PR. Thanks for working on #512; I definitely experienced those intermittent failures before you fixed it.

mislav:
Oh wait, it's because I disabled the multithread test in 134ea88. I'll try to re-enable it now that you've fixed it.

mislav:
@mrbrdo There seems to be [still a failure in `concurrency_test.rb`](https://travis-ci.org/vcr/vcr/jobs/85747586)

mrbrdo:
Can you try this fix https://github.com/mrbrdo/vcr/commit/fd6da084e99b35c7f4998e72953141d48810272e

mrbrdo:
It could be caused by this though:
https://github.com/vcr/vcr/blob/master/lib/vcr/middleware/excon.rb#L138
And then the error occurs here: https://github.com/vcr/vcr/blob/master/lib/vcr/structs.rb#L497

But then that's not 'my' fault :)

mislav:
It seems like you might be on the right track. But please do investigate further. You're much more familiar with this part of the codebase right now than I am.

In the meantime I feel that this PR can be merged, since it fixes rbx.

mrbrdo:
Yes I agree. Also don't think we need to merge that commit of mine I linked above, since it's quite possible it doesn't really fix anything.

@myronmarston https://github.com/vcr/vcr/commit/e25b014e42813d509ef97a4b24b5293f5cccb4a3#diff-6d39930288e6c17dd90c201056072c11R138 Could you look into why you added line 138 here (see above comments for reference)?

myronmarston:
> @myronmarston e25b014#diff-6d39930288e6c17dd90c201056072c11R138 Could you look into why you added line 138 here (see above comments for reference)?

I don't have any memories of writing that, but if you `git checkout e25b014`, `bundle install`, comment out line 138 and run `bundle exec rspec`, 15 excon tests fail, so it appears to be required due to how excon works. (In general, I only ever add lines like that if I can demonstrate the need through a test, so commenting out a line and seeing what tests break is always a good start with diagnosing the need for a line).  Anyhow, looking at the 15 failures, the backtrace shows it getting to that line via https://github.com/vcr/vcr/blob/e25b014e42813d509ef97a4b24b5293f5cccb4a3/lib/vcr/middleware/excon.rb#L37-L40 so it appears to have to do with a case when Excon is handling errors.  Hopefully that helps.

krainboltgreene:
What's the state of this? Can I click the button? :3

PR #514, opened 2015-10-30
Use the specs scope as the cassette name when the description is empty

tessi:
The scope remains relatively stable when new spec are inserted into the suite.
It's not as good as a description, but better than nothing.

Relates to (and possibly fixes) https://github.com/vcr/vcr/issues/378.
This patch helps a lot with `it { is_expected.to be 'something' }`-style specs.
Multiple of those specs overwrite each others cassettes, which is fixed by this patch.

https://github.com/vcr/vcr/issues/378 has a rather long discussion about which cassette name should be used in this case. I'd vote for the scope because
* we don't have a description
* the line number of the specs changes a lot when inserting new specs to the testsuite
* there is no easy way to get the source code of the spec (one could use the `method_source` gem, but it has other limitations for example with dynamically defined specs)

The nested scope does not change when changing existing specs, or appending specs to the suite.

### DIFF
--- a/lib/vcr/test_frameworks/rspec.rb
+++ b/lib/vcr/test_frameworks/rspec.rb
@@ -8,7 +8,12 @@ module VCR
       def configure!
         ::RSpec.configure do |config|
           vcr_cassette_name_for = lambda do |metadata|
-            description = metadata[:description]
+            description = if metadata[:description].empty?
+                            # we have an "it { is_expected.to be something }" block
+                            metadata[:scoped_id]
+                          else
+                            metadata[:description]
+                          end
             example_group = if metadata.key?(:example_group)
                               metadata[:example_group]
                             else
--- a/spec/lib/vcr/test_frameworks/rspec_spec.rb
+++ b/spec/lib/vcr/test_frameworks/rspec_spec.rb
@@ -17,6 +17,17 @@ describe VCR::RSpec::Metadata, :skip_vcr_reset do
         ])
       end
     end
+
+    context 'when the spec has no description' do
+      it do
+        expect(VCR.current_cassette.name.split('/')).to eq([
+          'VCR::RSpec::Metadata',
+          'an example group',
+          'when the spec has no description',
+          '1:1:2:1'
+        ])
+      end
+    end
   end
 
   context 'with the cassette name overridden at the example group level', :vcr => { :cassette_name => 'foo' } do


tessi:
I could use the Proc's source (instead of the scope) by using the `method_source` gem if you prefer that.

lib/vcr/test_frameworks/rspec.rb
```diff
@@ -8,7 +8,12 @@ module Metadata
       def configure!
         ::RSpec.configure do |config|
           vcr_cassette_name_for = lambda do |metadata|
-            description = metadata[:description]
+            description = if metadata[:description].empty?
+                            # we have an "it { is_expected.to be something }" block
+                            metadata[:scoped_id]
+                          else
+                            metadata[:description]
+                          end
```
krainboltgreene: Are we sure the state of things is that `metadata[:description]` will either be a full string `"something"` or an empty string `""`?

krainboltgreene:
Sorry this took so long to respond, I wasn't getting notifications and now that's fixed. Great catch, I'll merge once I'm sure about that question.

lib/vcr/test_frameworks/rspec.rb
tessi: That was the case while I had my debugger in that line. I can play save and add a nil-check, though :)

tessi:
@krainboltgreene thanks for answering :) Do you like the scope or the method_gem approach more? I can do either way.

lib/vcr/test_frameworks/rspec.rb
krainboltgreene: I think it's best, unless we can be better up the line and add a place that does metadata[:description] = ... || ""

krainboltgreene:
@tessi Scope is good enough. I'll keep a note for trying method_gem if people have problems.

lib/vcr/test_frameworks/rspec.rb
tessi: According to [this method](https://github.com/rspec/rspec-core/blob/19761e35a0fa8e7d61eaf4af52226fa0c2a39817/lib/rspec/core/metadata.rb#L173) in the rspec core, the description can never be nil.

tessi: @krainboltgreene is that info what you needed? Or shall I add a nil check?

krainboltgreene:
@tessi Thanks!

PR #516, opened 2015-11-02
make ignore_cassettes a boolean again to prevent nil.dup crashes

apachelogger:
According to the documentation turn_off's insert_cassettes is a boolean, if
it is not set as an option it will however be implicitly set to nil since
the thread-safety port. This in turn makes requests fail on nil.dup after
turn_off was called as dup_context would explicitly call dup for
ignore_cassettes.

To prevent this ignore_cassettes is now used as a boolean everywhere again
and defaults to false when turn_off! is called without ignore_cassettes
explicitly set. This should prevent crashes in dup_context.

### DIFF
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -262,7 +262,7 @@ module VCR
                                 "You must eject it before you can turn VCR off."
     end
 
-    set_context_value(:ignore_cassettes, options[:ignore_cassettes])
+    set_context_value(:ignore_cassettes, options.fetch(:ignore_cassettes, false))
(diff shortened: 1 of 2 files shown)


shekibobo:
Came across this issue today. My current workaround is to pass an empty array, but I don't know if there are other side-effects to doing that. 

```ruby
VCR.turn_off!(ignore_cassettes: [])
```

krainboltgreene:
:heart: :heart: :heart: :heart: :heart: :heart: :heart:

PR #522, opened 2015-11-16
Update upgrade doc, variable serializer does not exist [ci skip]

Edouard-chin:
My little contribution on the project :balloon:


krainboltgreene:
Thank you so much! That probably tripped up a lot of people.

Issue #517, opened 2015-11-03
Cucumber scenario outlines use incorrect cassette


AgentAntelope:
We're experiencing some odd behaviour since upgrading to `v3.0.0`, specifically we've tracked this down to https://github.com/vcr/vcr/commit/c16e8cc1b7066fc4e46c0c014ccd97509a038e87#diff-1c6bdcf6421dae582fbc386e7f988f5cR364.

We're using cucumber on a webrick server, which maintains its own thread for running the application. This means that in scenario outlines, the webrick thread duplicates the context (and thus, the cassette) of the main thread during the first scenario variant, but as the webrick thread isn't refreshed between tests this cassette is still in use in the subsequent scenario variants, even though the main thread's context has moved on.

The code that solved the issue for us changes `#get_context` like so: 

```@ruby
  def get_context(thread_key, fiber_key = nil)
    context = @context[fiber_key] if fiber_key
    context ||= @context[thread_key]
    if context
      context
    else
      @context[MainThread]
    end
  end
```

Essentially, we rely on the main thread to maintain our context for us, rather than duplicating that context to threads that may remain in use beyond the context of a specific scenario variant. However, I don't know whether this meets the requirements for thread safety and JRuby compatibility. I'd appreciate some feedback, and am happy to raise a PR if this solves the issue.

Thanks :)

bmulholland:
+1, I'm getting the same thing. In particular, what I'm seeing is that VCR isn't finding the recorded cassette. In the error, it tells me that it is looking in the cassette file for a previous, unrelated, cucumber test. I will downgrade cucumber for now.

djberg96:
Paging @mrbrdo - what do you think?

mrbrdo:
Good description, it makes sense.
I think there were a few reasons I duplicated the context instead of using the main thread's one. One is that it avoids locks so it's more performant, and I think it was also necessary because of `around_http_request`, since it will insert/eject cassettes for each request during the same test. Solving this in a better way might actually be much easier if https://github.com/vcr/vcr/issues/461 is done, as `around_http_request` is the source of much of the complexity here.

For this particular case: I am wondering if using your webrick's thread as the `MainThread` would solve the problem for you? You could try just overriding the constant to try this out, like:
```ruby
module ::VCR
  MainThread = some_thread
  initialize_ivars # important to call this as well
end
```
[...]

PR #529, opened 2015-12-06
Fix issue 517

mrbrdo:
Fix for #517 

I also included a commit with readme credits for the thread-safety work, mostly because I wanted to acknowledge Daniel Berger for putting up and paying out the bounty for that. Hope that's ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,6 +171,9 @@ Credits
     support.
   * [Jacob Green](https://github.com/Jacobkg) for help with ongoing VCR
     maintenance.
+  * [Jan Berdajs](https://github.com/mrbrdo) and [Daniel Berger](https://github.com/djberg96)
+    for improvements to thread-safety.
+
 
 Thanks also to the following people who have contributed patches or helpful suggestions:
 
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -4,6 +4,7 @@ require 'vcr/util/variable_args_block_caller'
 require 'vcr/cassette'
 require 'vcr/cassette/serializers'
 require 'vcr/cassette/persisters'
+require 'vcr/linked_cassette'
 require 'vcr/configuration'
 require 'vcr/deprecations'
 require 'vcr/errors'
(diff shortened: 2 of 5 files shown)


djberg96:
@mrbrdo Thanks!

Seems these travis failures are not the result of your code. Rebase maybe?

mcfiredrill:
The test suite looks to be passing, the `script/ci.sh` is failing at the end of the build for some reason though.
```
script/ci.sh: line 103: printf: `%': missing format character
The command "script/ci.sh" exited with 1.
```
Seeing the same behaviour on all ruby versions.

mrbrdo:
Yes I noticed this too. The tests were failing before (the tests, not this printf error), and I force pushed a few times to fix those errors. So not sure what this is. Perhaps you just need to re-run the build on Travis?

mcfiredrill:
Oh lol....I see the issue now.
Check out line 103 of `script/ci.sh`
https://github.com/vcr/vcr/blob/master/script/ci.sh#L103
```
printf "Failed: documentation coverage is less than 100%"
```

This is indeed missing the format character. The `%` should be escaped. Or its probably fine to just use echo.

Regardless, it looks like this is intended to fail the build anyway if documentation is less than 100%. As `STATUS` is set to 1 and `exit $STATUS` is called.
https://github.com/vcr/vcr/blob/master/script/ci.sh#L104

mrbrdo:
yes definitely, will fix the documentation.

mcfiredrill:
@mrbrdo I think so. Looks like you can run `bundle exec yard stats --list-undoc` to list the undocumented code.

mrbrdo:
Also I knew I remembered something about this printf error... I already took a stab at it when doing #512 :) https://github.com/mrbrdo/vcr/commit/97a4ffc8507c642ce652cf5adf53ffbd5a068e7c
I guess I missed the percent issue though. But it seems the color codes weren't working either.

mrbrdo:
It seems the build is passing now, not sure why it's not linked here...
https://travis-ci.org/vcr/vcr/builds/95870650

mcfiredrill:
:+1:

djberg96:
Excellent work. Please merge.

PR #533, opened 2015-12-09
script/ci.sh: use echo instead of printf if doc coverage test failed

mcfiredrill:
printf errors if you don't escape the `%` symbol:

```
$ printf "Failed: documentation coverage is less than 100%"
-bash: printf: `%': missing format character
```

It seems fine to just use echo here instead.


mislav:
:+1:

PR #539, opened 2015-12-22
remove .ruby-version

mcfiredrill:
VCR should work on all currently maintained ruby versions. I propose we
delete this file.


krainboltgreene:
While it should, there's a lot of barrier to entry that this file removes for new people who want to contribute.

mcfiredrill:
Can you explain what you mean? I'd say this file *is* a barrier to entry, rather than removing barriers. They might not have the exact version installed.

mislav:
:+1: 

A `.ruby-version` file is not needed for a project that will work with any Ruby version. It just **adds a barrier to entry** because now people have to install a specific Ruby version, even though they probably already had a perfectly valid Ruby to start developing with.

krainboltgreene:
I have to disagree here. I've seen plenty of new developers not able to really figure out what to do at start and the `.ruby-version` file gives them a clear indication on how to start. If they've got rvm installed (highly likely) it *literally gives them the command to run*.

That said, I'll abide by the votes if it gets a second plus one.

mislav:
I can't imagine someone new to Ruby development (and thus having no Ruby versions installed) picking up VCR and starting hacking on it. VCR is a fairly advanced piece of software. I would guess that whoever plays with it already has a development environment set up.

mcfiredrill:
:+1:

shaicoleman:
I think it's better to remove it. It forces using an arbitrary and specific old version of Ruby.

Standard practice is not to include a .ruby-version, and add it in .gitignore for libraries meant to run on multiple versions.

Example: https://github.com/github/gitignore/blob/master/Ruby.gitignore

mcfiredrill:
@shaicoleman :+1:

mcfiredrill:
@krainboltgreene  I mean on this PR for example, looks like the status check from LGTM is still here.

Not sure if we can remove it without starting a new PR... :sweat:

krainboltgreene:
@mcfiredrill Tada!

mcfiredrill:
@krainboltgreene wow, :sparkles: :heart_eyes_cat:

mcfiredrill:
@krainboltgreene I don't know what you did but thank you! :green_heart:

PR #540, opened 2015-12-22
we use rspec 3 now

mcfiredrill:
I don't think we need to specify which version in this file.
:sweat_smile:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ You should be able to run the tests now:
 bundle exec rake
 ```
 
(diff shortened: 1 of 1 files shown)


krainboltgreene:
:+1:

mcfiredrill:
:+1:

mcfiredrill:
@krainboltgreene I'm having the same issue here. There are two +1's but I cannot merge the PR.
I'm not sure, but maybe it has to do with rebasing and force pushing again. It seems like the status is not updated after I update the branch.

PR #543, opened 2016-01-03
Warning: Do not use test/fixtures, fix #527

### LINKED ISSUE #527: Add warning - do *NOT* put cassettes in test/fixtures if using Rails minitest
The documentation should prominently warn people to *NOT* put vcr_cassettes inside the 'test/fixtures' directory when using Rails and minitest.

It *seems* logical to do this, but as noted in <http://guides.rubyonrails.org/testing.html>, "Rails by default automatically loads all fixtures from the test/fixtures folder for your models and controllers test."  The result: The system will try to load the vcr cassettes as models, causing mysterious error messages.

I decided to use 'test/vcr_cassettes/' which works fine (because I wanted all test materials inside test/).

A little warning might save debugging time for someone else...!

david-a-wheeler:
The directory 'test/fixtures' *looks* like a reasonable place to put VCR cassettes,
but if you use Rails + minitest the result is a mess (Rails will try to load these as models) -
with very confusing error messages.

Warn people to NOT do this, so that they won't accidentally fall into this trap.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ class VCRTest < Test::Unit::TestCase
 end
 ```
 
-Run this test once, and VCR will record the HTTP request to `fixtures/vcr_cassettes/synopsis.yml`.  Run it again, and VCR will replay the response from iana.org when the HTTP request is made.  This test is now fast (no real HTTP requests are made anymore), deterministic (the test will continue to pass, even if you are offline, or iana.org goes down for maintenance) and accurate (the response will contain the same headers and body you get from a real request).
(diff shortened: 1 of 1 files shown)


krainboltgreene:
Might be worth just having a FAQ section with this warning, but otherwise :+1:. Thanks for the quality of life improvement.

david-a-wheeler:
The only problem with a FAQ entry is that you need to know to look at the FAQ entry.

Changing the directory setting 'to taste' seems like such an obvious and easy change that a reader might not even *look* at the FAQ.

krainboltgreene:
Agreed.

On Sun, Jan 3, 2016 at 10:48 AM, David A. Wheeler <notifications@github.com>
wrote:

> The only problem with a FAQ entry is that you need to know to look at the
> FAQ entry.
>
> Changing the directory setting 'to taste' seems like such an obvious and
> easy change that a reader might not even *look* at the FAQ.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/vcr/vcr/pull/543#issuecomment-168528180>.
>



-- 
Kurtis Rainbolt-Greene, Hacker
Software Developer
1631 8th St.
New Orleans, LA
70115

mcfiredrill:
LGTM :+1:

PR #544, opened 2016-01-05
Require ruby 1.9.3

knu:
VCR 3.0.1 already does not work with ruby 1.8.7 due to dependency on Fiber, so it should be declared both in gemspec and in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,11 +96,10 @@ VCR has been tested on the following ruby interpreters:
   * MRI 2.1
   * MRI 2.2
   * MRI 2.3.0
-  * REE 1.8.7
   * JRuby
   * Rubinius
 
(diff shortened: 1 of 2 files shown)


mcfiredrill:
Thanks. :+1: LGTM
Thanks for noting we dropped 1.9.2 support as well.

mcfiredrill:
@knu Would you mind rebasing? I updated the list of supported rubies in another commit. 
https://github.com/vcr/vcr/pull/549x

knu:
@mcfiredrill Done.

mcfiredrill:
@knu thank you!
@krainboltgreene does this look OK to you?

krainboltgreene:
Yep! LGTM!

PR #546, opened 2016-01-07
fix webmock thread safety for with_global_hook_disabled

### LINKED ISSUE #545: There appears to be a bug in WebMock's after_request hook...
I am still getting this warning sometimes, and it appears as though the request bypasses vcr (i.e. it actually does the request).

Any idea what the status of this is and how to debug? I am using the latest webmock version (1.22.3), and although it seems the related PRs were merged, this bug is still coming up. It could be related to threads in my case, but I would like to figure this out.

For now I can see that `@__typed_vcr_request` was not there at the first invocation of `typed_request_for` for this specific request.

mrbrdo:
`with_global_hook_disabled` was not thread-safe. also now done in a more reliable way (only disable for specific request). since `with_global_hook_disabled` was marked as private API, and doing a google search did not reveal anyone using this method, I did not make it backwards-compatible.

Also I did the thread hash a bit more properly here, by defining a thread finalizer to remove the key. Did not do this in my main VCR thread-safety commits. However I don't feel it is a real issue due to the fact VCR does not run in long-living processes, so the extra memory used by holding information for dead threads is probably not a problem.

Fixes #545

### DIFF
--- a/lib/vcr/library_hooks/webmock.rb
+++ b/lib/vcr/library_hooks/webmock.rb
@@ -10,17 +10,32 @@ module VCR
     module WebMock
       extend self
 
-      attr_accessor :global_hook_disabled
-      alias global_hook_disabled? global_hook_disabled
+      @global_hook_disabled_requests = {}
 
-      def with_global_hook_disabled
-        self.global_hook_disabled = true
+      def with_global_hook_disabled(request)
+        global_hook_disabled_requests << request
 
         begin
           yield
         ensure
-          self.global_hook_disabled = false
+          global_hook_disabled_requests.delete(request)
         end
+      end
+
+      def global_hook_disabled?(request)
+        requests = @global_hook_disabled_requests[Thread.current.object_id]
+        requests && requests.include?(request)
+      end
+
+      def global_hook_disabled_requests
+        requests = @global_hook_disabled_requests[Thread.current.object_id]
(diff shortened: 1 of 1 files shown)


mcfiredrill:
@mrbrdo Sorry for no response to this for awhile. 
Are the build failures related?

mrbrdo:
Sorry, they are. I forgot to recheck the tests after I made the last changes. Should be fixed now.
Funnily enough, I had no errors while using this branch in one of my projects. Also, so far it seems this finally completely fixed my vcr-related thread-safety issues, since I am using webmock.

mrbrdo:
The remaining failure doesn't seem to be caused by my changes.

mcfiredrill:
The failure is indeed unrelated. (I'm hoping to fix that failure here: https://github.com/vcr/vcr/pull/551 )
:+1: 

Looking for another :+1:  or LGTM for this PR.....
Have you had a chance to look at this yet @krainboltgreene ?

krainboltgreene:
We'll need to have this be updated, but otherwise :+1:

mcfiredrill:
Yep, @mrbrdo mind rebasing? Sorry for the delay again...

mrbrdo:
there ya go

mcfiredrill:
@mrbrdo Thanks again!

Issue #478, opened 2015-03-05
Improve the debug message of VCR


dnesteryuk:
When there isn't a cassette to handle a request, VCR returns an error like:

```
An HTTP request has been made that VCR does not know how to handle:
  GET http://google.com/client/api?listAll=true&page=2&pagesize=20&response=json
  Body: 

VCR is currently using the following cassette:
  - /home/dnesteryuk/project/cars.yml
  - :record => :none
  - :match_requests_on => [:method, #<struct VCR::RequestMatcherRegistry::URIWithoutParamsMatcher params_to_ignore=["_", "sessionkey"]>, :body]

Under the current configuration VCR can not find a suitable HTTP interaction
to replay and is prevented from recording new requests. There are a few ways
you can deal with this:

  * If you're surprised VCR is raising this error
    and want insight about how VCR attempted to handle the request,
    you can use the debug_logger configuration option to log more details [1].
  * You can use the :new_episodes record mode to allow VCR to
    record this new request to the existing cassette [2].
  * If you want VCR to ignore this request (and others like it), you can
    set an `ignore_request` callback [3].
  * The current record mode (:none) does not allow requests to be recorded. You
    can temporarily change the record mode to :once, delete the cassette file 
    and re-run your tests to allow the cassette to be recorded with this request [4].
```

When a lot of cassettes are applied:

```ruby
  VCR.use_cassette('cars') do
    VCR.use_cassette('cars2') do
      VCR.use_cassette('cars3') do
        VCR.use_cassette('cars4') do
          # do HTTP call here
        end
      end
    end
  end
```

This message doesn't help too much. I was thinking to improve that message with all lists of aplied cassettes:

```
An HTTP request has been made that VCR does not know how to handle:
  GET http://google.com/client/api?listAll=true&page=2&pagesize=20&response=json
  Body: 

VCR is currently using the following cassettes:
  - /home/dnesteryuk/project/cars.yml
  - /home/dnesteryuk/project/cars2.yml
  - /home/dnesteryuk/project/cars3.yml
  - /home/dnesteryuk/project/cars4.yml

Under the current configuration VCR can not find a suitable HTTP interaction
to replay and is prevented from recording new requests. There are a few ways
you can deal with this:

  * If you're surprised VCR is raising this error
    and want insight about how VCR attempted to handle the request,
    you can use the debug_logger configuration option to log more details [1].
  * You can use the :new_episodes record mode to allow VCR to
    record this new request to the existing cassette [2].
[...]

PR #501, opened 2015-08-13
PR to improve the debug message (Unhanded request error)

dnesteryuk:
This PR allows us to see names and parameters of all cassettes in the stack. Example:

```
================================================================================
       An HTTP request has been made that VCR does not know how to handle:
         GET http://google.com/client/api?listAll=true&page=2&pagesize=20&response=json
  Body:
       
       VCR are currently using the following cassettes:
         - /home/dnesteryuk/projects/vcr/tmp/cassette_library_dir/sample.yml
           - :record => :once
           - :match_requests_on => [:method, :uri]
         - /home/dnesteryuk/projects/vcr/tmp/cassette_library_dir/example.yml
           - :record => :new_episodes
           - :match_requests_on => [:method, :uri]
       
       Under the current configuration VCR can not find a suitable HTTP interaction
       to replay and is prevented from recording new requests. There are a few ways
       you can deal with this:
       
         * If you're surprised VCR is raising this error
           and want insight about how VCR attempted to handle the request,
           you can use the debug_logger configuration option to log more details [1].
         * You can use the :new_episodes record mode to allow VCR to
           record this new request to the existing cassette [2].
         * If you want VCR to ignore this request (and others like it), you can
           set an `ignore_request` callback [3].
       
       [1] https://www.relishapp.com/vcr/vcr/v/2-9-3/docs/configuration/debug-logging
       [2] https://www.relishapp.com/vcr/vcr/v/2-9-3/docs/record-modes/new-episodes
       [3] https://www.relishapp.com/vcr/vcr/v/2-9-3/docs/configuration/ignore-request
       ================================================================================
```

It gives better understanding what is used for stubbing HTTP requests. This change was discussed here #478.

PR #555, opened 2016-02-25
Don't eject cassettes on UnhandledHTTPRequestError

hdabrows:
For context see the commit message and https://github.com/vcr/vcr/issues/478#issuecomment-130683370.

### DIFF
--- a/lib/vcr/errors.rb
+++ b/lib/vcr/errors.rb
@@ -88,18 +88,7 @@ module VCR
       end
 
       def current_cassettes
-        @cassettes ||= begin
-          cassettes = VCR.cassettes.to_a.reverse
-
-          begin
-            loop do
-              break unless VCR.eject_cassette
-            end
-          rescue EjectLinkedCassetteError
-          end
-
-          cassettes
-        end
+        @cassettes ||= VCR.cassettes.to_a.reverse
       end
 
       def request_description


mcfiredrill:
Sorry I'm kind of confused about what issue this is solving exactly.
In the thread you linked I closed the issue because https://github.com/vcr/vcr/pull/501 was merged.
Does https://github.com/vcr/vcr/pull/501 solve your issue?

hdabrows:
```ruby
#!/usr/bin/env ruby

require "bundler/inline"
gemfile true do
  source "https://rubygems.org"

  gem "minitest"
  gem "vcr", "3.0.3"
  gem "webmock"
end

include WebMock::API
WebMock.enable!
WebMock.disable_net_connect!

VCR.configure do |c|
  c.cassette_library_dir = "."
  c.hook_into :webmock
end

require "minitest/autorun"
class BugTest < Minitest::Test
  def setup
    VCR.insert_cassette("cassette")
  end

  def test_bug
    begin
      Net::HTTP.get("example.com", "/#{SecureRandom.hex}")
    rescue => e
      # this will rescue VCR::Errors::UnhandledHTTPRequestError
    end
  end

  def teardown
    Net::HTTP.get("example.com", "/")
    VCR.eject_cassette
  end
end
```

Please run twice in the same directory (second run will use the cassette). Then try with `3.0.0`.

#501 included the code that ejects the cassettes on error. The reason for this was explained in https://github.com/vcr/vcr/issues/478#issuecomment-130683370. At that time the only way to get the list of all cassettes was to call a private method on `VCR` module. This is no longer the case after https://github.com/vcr/vcr/pull/529 exposed that method as private. I assume the reason why #529 preserved this behaviour (ejecting cassettes) is because the person writing the code didn't know that it was no longer necessary to do so.

mcfiredrill:
@hdabrows Thanks alot for taking the time to write up the testcase. 💚 

I am continuing to investigate this issue, and it really helps to have this simple stripped down example.

XrXr:
Any update on this PR? I think it also fixes #659

krainboltgreene:
Lets get this updated from master so we can get it merged in.

hdabrows:
@krainboltgreene I've rebased the branch on master.

PR #557, opened 2016-03-06
Updating the README to support Betamax Java (which is no longer Groovy).

cowboygneox:
Betamax (Groovy) is no longer Groovy, it's now Java. Also, Betamax has been revived and is in active development. Would there be objection to moving it higher on the list?


mcfiredrill:
LGTM

krainboltgreene:
:+1:

PR #567, opened 2016-03-26
Re-record is opt-in; made that clearer

perlun:
The previous wording implied that re-recording is _automatic_, i.e. enabled by default. The fact is that it isn't, as stated by the Rubydocs for the `use_cassette` method: http://www.rubydoc.info/gems/vcr/VCR#use_cassette-instance_method

The purpose of this PR is to try and make this more apparent here for people like me, who are new to the VCR project. I'm open for suggestions if you feel that the wording here isn't perfect; the goal is simply to make it clear enough that everyone understands what we mean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Run this test once, and VCR will record the HTTP request to `fixtures/vcr_casset
   * The same request can receive different responses in different tests--just use different cassettes.
   * The recorded requests and responses are stored on disk in a serialization format of your choice (currently YAML and JSON are built in, and you can easily implement your own custom serializer) and can easily be inspected and edited.
   * Dynamic responses are supported using ERB.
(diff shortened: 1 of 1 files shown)


mcfiredrill:
This is actually referring to a feature:
https://relishapp.com/vcr/vcr/v/1-6-0/docs/cassettes/automatic-re-recording

perlun:
Yes, I know. But `re_record_interval` defaults to "never".

To explain my point: I started using the project yesterday, and because it said "Automatically re-records cassettes" I started looking for how to disable it (because it was needed in one of my use cases). It took some digging, and then I found that it is opt-in, not enabled by default. I wanted to make that clearer, so that other people won't be confused like I was.

mcfiredrill:
I see what you're saying now, that makes sense to me. :+1:

PR #571, opened 2016-04-21
Upgrade to latest version of rake, cucumber, and aruba

kulte:


kulte:
It looks like this upgrade required some deprecation warnings to be fixed. I'm doing that as I run the Cucumber specs locally, but the upgrade also led to some `Cucumber::Ambiguous` exceptions being thrown that aren't being thrown with the current versions of the dependencies. Does anyone have more knowledge of Cucumber than I do and is able to point me in the right direction?

mcfiredrill:
Closing as this PR is a bit stale but really I'd prefer a separate PR for each Gem rather than upgrading everything in the same PR.

PR #575, opened 2016-05-06
Added Flickr's yakbak in list of JS/Node ports

dannguyen:
Announcement post here: http://code.flickr.net/2016/04/25/introducing-yakbak-record-and-playback-http-interactions-in-nodejs/


mcfiredrill:
Thanks

PR #579, opened 2016-05-13
enable webmock explicitly

mcfiredrill:
Starting with webmock 2.0, `WebMock.enable!` needs to be called
explicitly. We should do this for users automatically when they
configure VCR to use webmock.

This seems to be backwards compatible with 1.x versions of webmock.


lib/vcr/library_hooks/webmock.rb
```diff
@@ -4,6 +4,8 @@
 
 VCR::VersionChecker.new('WebMock', WebMock.version, '1.8.0').check_version!
```
mcfiredrill: Hmm...should this change?

PR #585, opened 2016-05-30
allow non lantin characters in filename

0x01f7:


mcfiredrill:
This looks good, and tests pass. 
I can't see how it would break anything else.

krainboltgreene:
Famous last words.

On Wed, Jun 8, 2016 at 10:19 AM, Tony Miller <notifications@github.com>
wrote:

> This looks good, and tests pass.
> I can't see how it would break anything else.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/vcr/vcr/pull/585#issuecomment-224664059>, or mute the
> thread
> <https://github.com/notifications/unsubscribe/AAUb2ZyZ0It4n-H-rjq5-NKueFyb91OBks5qJvm2gaJpZM4Ip36a>
> .
>



-- 
Kurtis Rainbolt-Greene, Software Developer
Founder of Difference Engineers

2985 San Marino St.
Los Angeles, CA 90006
202-643-2263

0x01f7:
@mcfiredrill @krainboltgreene 

Not really, it will cause a compatibility issue for people who already use non lantin chars in their testcases, they need to modify the filename of the exist cassettes, or re-records cassettes.

But in my opinion, the purpose of `sanitized_file_name_from` is to provide a valid and recognizable filename, which could be used in any filesystem. So why not support non lantin chars?

0x01f7:
@mcfiredrill @kirkconnell 

Is anyone interested in this feature?

Since it may cause a compatibility issue, maybe we should make it configurable, provide a signleton method like [carrierwave#filenames-and-unicode-chars]?

Let me know if there is anything else I can do for it.


[carrierwave#filenames-and-unicode-chars]:https://github.com/carrierwaveuploader/carrierwave#filenames-and-unicode-chars

ecnalyr:
> Is anyone interested in this feature?

I like it.  Add a note about the possible compatibility issue for people who previously used non-latin chars in their test cases and merge it.

krainboltgreene:
I definitely want this merged.

krainboltgreene:
@0x01f7 Can you give me an example of a break in expectation?

0x01f7:
@krainboltgreene 

> Can you give me an example of a break in expectation?

```ruby
require 'rubygems'
require 'test/unit'
require 'vcr'

VCR.configure do |config|
  config.cassette_library_dir = "fixtures/vcr_cassettes"
  config.hook_into :webmock # or :fakeweb
end

class VCRTest < Test::Unit::TestCase
  def test_example_dot_com
    VCR.use_cassette("摘要") do # use non latin characters, like Simplified Chinese
      response = Net::HTTP.get_response(URI('http://www.iana.org/domains/reserved'))
      assert_match /Example domains/, response.body
    end
  end
end
```

Without this feature, VCR will record the request to `fixtures/vcr_cassettes/_.yml`. But now it will record the request to `fixtures/vcr_cassettes/摘要.yml`.

People who already use non latin chars in their testcases, they need to modify the filename of the exist cassettes(`_.yml` -> `摘要.yml`), or re-records cassettes.

krainboltgreene:
Ahhh, that's what I figured. I wonder if most peopl,e deal with it or hack vcr to make it work?

Either way, @mcfiredrill this should definitely go in and major version bump.

0x01f7:
> I wonder if most peopl,e deal with it or hack vcr to make it work?

Recently I joined an existed project which use whole Simplified Chinese words to describe the test examples, for example:

```ruby
# something
describe '某物' do
  # do something
  it '做一些事' do
  end

  # do antoher thing
  it '做另一些事' do
  end
end 
```

If you run this example, **ONLY ONE FILE** named `fixtures/vcr_cassettes/_/_.yml` will generated, now you have to deal with it. The monkey patch below is used to make it work properly in our project.

```ruby
module VCR
  class Cassette
    class Persisters
      # The only built-in cassette persister. Persists cassettes to the file system.
      module FileSystem
        def sanitized_file_name_from(file_name)
          parts = file_name.to_s.split('.')

          if parts.size > 1 && !parts.last.include?(File::SEPARATOR)
            file_extension = '.' + parts.pop
          end

          # HACK!
          parts.join('.') + file_extension.to_s
        end
      end
    end
  end
end
```

As far as I know, another workaround is define a custom cassette persister, then configure :persist_with option. But it's not easy for people to find it.

In my opinion, the best solution is make a built-in support. Thx. :-)

mcfiredrill:
@krainboltgreene I'll try to find some time to test this, and seems like a major version bump is the only way. 👌

PR #591, opened 2016-06-15
Remove trailing whitespaces from YAML files

exAspArk:
Hi, 

I very often deal with YAML cassettes that contain trailing whitespaces. It's quite annoying, especially if your editor removes trailing whitespaces in YAML files automatically 😸 

The reason why it happens that Ruby serializers such as `YAML`, `Psych` use empty values by default for `nil`. However, the latest YAML [specification v1.2](http://www.yaml.org/spec/1.2/spec.html#id2803362) says that `null` is a canonical form to describe it (in the previous [specification v1.1](http://yaml.org/type/null.html) `~` was a canonical form).

To get rid of these trailing whitespaces and prevent issues like #466 in the future, I would suggest replacing them with `null` values. 

P.S. I've tried to use `canonical: true` option while dumping, but I don't think we really need to use a fully canonical YAML version. Compare: 

```
# default     |   # fully canonical
---           |   ---
a: 7          |   !!map {
null value:   |     ? !!str "a"
nested:       |     : !!int "7",
  hash:       |     ? !!str "nested"
    - 1       |     : !!map {
    - 2       |       ? !!str "hash"
    - 3       |       : !!seq [
              |         !!int "1",
              |         !!int "2",
              |         !!int "3",
              |       ],
              |     },
              |     ? !!str "null value"
              |     : !!null "null",
              |   }
```


krainboltgreene:
I totally get this, but it feels really hacky. I wonder if there's middle ground with the canonical.

Otherwise, fuck it, lets do canonical.

krainboltgreene:
Mind updating from master?

krainboltgreene:

exAspArk:
@krainboltgreene hey, I updated the PR against the latest master

olleolleolle:
@exAspArk Since it's been a season or two, and we were busy with other things, could you  update from master?

(Just say "no" if you'd rather that we carry the commit in another PR.)

PR #597, opened 2016-07-07
Fix failing Excon tests

penman:
Excon 0.50.0 renamed `Excon::Errors::Error` to `Excon::Error`.
For backwards compatibility, `Excon::Errors::Error` was redefined as a subclass of `Excon::Error`, but this doesn't help when testing the class of an error raised by Excon.

I've opened an issue on Excon (excon/excon#580) to ask for this to be changed from a subclass to an assignment, which would prevent these failures, but in the meantime, this PR fixes the tests.


penman:
This is fixed in Excon now.

mcfiredrill:
@penman Thanks!

PR #598, opened 2016-07-15
Allow to continue to play a cassette after errors

exAspArk:
Hi, 

I used VCR version `3.0.0` in my application. In one of my tests I made two requests:
- The first one was allowed to fail for any reason. It means that even if it wasn't recorded and failed by VCR ("does not know how to handle") the application should continue working.
- The second request was recorded, so the test should successfully pass.

However, after upgrading it to `3.0.1` the second request didn't play the cassette.

After some bisecting between these 2 versions I found this commit b261b7a. 

The problem is that generating a message for an error ejects cassette. So I would suggest keeping cassettes even after raising a VCR error.

### DIFF
--- a/lib/vcr/errors.rb
+++ b/lib/vcr/errors.rb
@@ -88,18 +88,7 @@ module VCR
       end
 
       def current_cassettes
-        @cassettes ||= begin
-          cassettes = VCR.cassettes.to_a.reverse
-
-          begin
-            loop do
-              break unless VCR.eject_cassette
-            end
-          rescue EjectLinkedCassetteError
-          end
-
-          cassettes
-        end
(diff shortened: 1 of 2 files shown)


mcfiredrill:
@exAspArk is this related?
https://github.com/vcr/vcr/pull/555

exAspArk:
@mcfiredrill oh yeah, thanks! Haven't seen it, I guess it is the same issue.

My test case shows that after failing on the first request (`(:once) does not allow new requests to be recorded`) it will continue playing the same cassette without raising an error `There is currently no cassette in use`.

krainboltgreene:

exAspArk:
It looks like it was fixed in this PR https://github.com/vcr/vcr/pull/555. Going to close it

PR #624, opened 2016-12-07
Include Cucumber test suite in the default Rake task.

ecnalyr:
Run RSpec and Cucumber suites by default when `bundle exec rake` is run from the console, _not just_ RSpec.

- Previously we only ran RSpec, yet we have a reasonably sized Cucumber suite that could get overlooked by contributors.
- Our [CONTRIBUTING.md file](https://github.com/ecnalyr/vcr/blob/master/CONTRIBUTING.md#L10-L14) suggests the default rake task runs our full test suite, this change helps reinforce that idea.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -11,7 +11,7 @@ task default: :spec
 
 using_git = File.exist?(File.expand_path('../.git/', __FILE__))
 
-require 'cucumber/rake/task'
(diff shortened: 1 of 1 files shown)


mcfiredrill:
Nice! Will merge once CI passes.

PR #625, opened 2016-12-07
Allow rbx-2 to fail without breaking Travis build.

ecnalyr:
- The rbx-2 buils fails most of the time on Travis, but not because of a problem with VCR’s codebase.
- Precedence?  Capistrano recently made the same change: https://github.com/capistrano/capistrano/pull/1784

I feel like the red Travis builds are making it _not fun_ to contribute to this project, trying to get our builds green.

This can be undone once rvm releases a new version that includes the fix merged into master that was discussed here: https://github.com/rvm/rvm/issues/3617

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,3 +12,4 @@ rvm:
 matrix:
   allow_failures:
     - rvm: jruby-19mode
+    - rvm: rbx-2


ecnalyr:
An alternative to allowing failure would be to correct the failure by using a different version of rvm as seen here: https://github.com/vcr/vcr/pull/626

mcfiredrill:
Actually I'm pretty sure this is what broke the rbx-2 build:
https://github.com/vcr/vcr/pull/619

I could try reverting it, but honestly I don't really mind allowing the failure either as this is just kind of another obstacle to contributing in my mind. :\

ecnalyr:
Nah, the changes in #619 pr did not cause this problem.

Lots of people are having this issue with rbx-2 and travis right now (for a few months now, actually e.g. https://github.com/vmg/redcarpet/issues/580).

mcfiredrill:
@ecnalyr Thanks for letting me know.  I will merge this for now, at least we will have some green builds going forward.

PR #627, opened 2016-12-09
revert removing .ruby-gemset

mcfiredrill:
See if this fixes the rbx build.


mcfiredrill:
Thought I saw some rbx builds pass before I removed the .ruby-gemset file, but looks like there is a failure anyway.

Issue #578, opened 2016-05-13
version 4.0


mcfiredrill:
This is an issue to track the major breaking changes happening in version 4.0.

- [x] remove fakeweb support
- [x] remove VCR::RSpec::Macros https://github.com/vcr/vcr/issues/180#issuecomment-8391651
- [x] drop ruby 1.9.3 support

mcfiredrill:
- [ ] allow non lantin characters in filename  https://github.com/vcr/vcr/pull/585

ecnalyr:
> remove VCR::RSpec::Macros 

Do you mean to completely remove the [already-deprecated](https://github.com/vcr/vcr/pull/224) `Macros` module, or is there more stuff that needs to be deprecated somewhere else first?

I don't think there is anything left of the `Macros` module that isn't deprecated, but I wanted to be sure I understood.

mcfiredrill:
@ecnalyr It looks like VCR::RSpec::Macros is already in lib/vcr/deprecations.rb with the appropriate deprecation warnings being issued, so I don't think there is anything else that should be deprecated before we remove it in 4.0. 

Is there anything else related to Macros you thought should be deprecated before we remove it?

ecnalyr:
> Is there anything else related to Macros you thought should be deprecated before we remove it?

No, was just wanting to make sure I understood the scope.

mcfiredrill:
PR to remove fakeweb
https://github.com/vcr/vcr/pull/640

paul:
Dropping fakeweb is pretty much required to work in Ruby 2.4.x. How is progress towards those other two tasks? Any chance of a release without them?

mcfiredrill:
Should we remove typhoeus support as well?
https://github.com/vcr/vcr/issues/464#issuecomment-68910577

mcfiredrill:
Hi its been awhile but I want to release 4.0 soon if that's cool, because I want to use https://github.com/vcr/vcr/pull/585

krainboltgreene:
Ready when you are.

mcfiredrill:
Released 4.0.0 this morning 🌞

PR #629, opened 2017-01-17
[WIP] add ruby 2.4.0 to travis test matrix

mcfiredrill:


bdmac:
@mcfiredrill so I forked this intending to help fix vcr compatibility with Ruby 2.4 but I've hit a bit of a snag...

All of the errors happening in vcr are actually errors in fakeweb relating to the way it is trying to mock out parts of Net::Http. Specifically there are two changes needed in fakeweb to get things working again:

1. Its monkeypatch of Net::BufferedIO has an initializer alias method chain that no longer matches the initializer of this class in 2.4. This is the root cause of all the `undefined method '<<' for {:read_timeout=>60, :continue_timeout=>nil, :debug_output=>nil}:Hash` failures in vcr. Updating the initializer override to match the new method signature and then continue to do its stubbing fixed this error.
2. After fixing that I found that fakeweb also has a `FakeWeb::StubSocket` class that is missing a stub of the (new?) `close` method that gets called by the Net::Http in 2.4. I had to add a stub `def close; @closed = true; end` basically.

Those changes make the specs in vcr pass but from what I can see nobody's maintaining fakeweb anymore so I have no idea how you guys want to proceed...

There are also a few deprecation warnings related to the merging of `Fixnum` with `Integer` in Ruby 2.4.  The diff to fix that looks like this but probably breaks < 2.4. It would need some minor version checks I guess?

```
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: spec/lib/vcr/version_spec.rb
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ version_spec.rb:15 @ describe "VCR.version" do

  describe '#major' do
    subject { super().major }
    it { should be_instance_of(Fixnum) }
    it { should be_instance_of(Integer) }
  end

  describe '#minor' do
    subject { super().minor }
    it { should be_instance_of(Fixnum) }
    it { should be_instance_of(Integer) }
  end

  describe '#patch' do
    subject { super().patch }
    it { should be_instance_of(Fixnum) }
    it { should be_instance_of(Integer) }
  end
end
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: spec/support/fixnum_extension.rb
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ fixnum_extension.rb:2 @
unless 7.respond_to?(:days)
  class Fixnum
  class Integer
    def days
      self *
      24 * # hours
```

bdmac:
@mcfiredrill I'd push up my `vcr` branch but again all the fixes are actually to fakeweb.

mcfiredrill:
@bdmac Thank you so much for putting in the time on this. 
Wondering what we can do about our dependency on fakeweb, looks like its been unmaintained for a few years. At least its good to hear most of the issues are coming from something we can possibly remove. 😅 

As for the warnings, I fixed them here https://github.com/vcr/vcr/pull/630 and it seems to be backwards compatible. I guess this PR needs a rebase against master.

mcfiredrill:
Since we plan to drop fakeweb support in 4.0, I wonder if we should bother getting it to work with 2.4.
https://github.com/vcr/vcr/issues/578

mcfiredrill:
Only thing I can think of is, if your monkeypatch is relatively simple and backwards compatible perhaps we could include it in the 3.x series and remove it in 4.0. 🤔

0x01f7:
This is an issue with Fakeweb rather than VCR. I think we should not include any monkeypatch in VCR library. I open a [PR](https://github.com/vcr/vcr/pull/640) to drop fakeweb support. :-)

bdmac:
@mcfiredrill so you'd have to monkeypatch two things:

lib/fake_web/ext/net_http.rb (update init override with correct 2.4 signature)
```ruby
module Net
  class BufferedIO
    def initialize_with_fakeweb(io, read_timeout: 60, continue_timeout: nil, debug_output: nil)
      @io = io
      @read_timeout = read_timeout
      @continue_timeout = continue_timeout
      @debug_output = debug_output
      @rbuf = ''

      @io = case io
      when Socket, OpenSSL::SSL::SSLSocket, IO
        io
      when String
        if !io.include?("\0") && File.exists?(io) && !File.directory?(io)
          File.open(io, "r")
        else
          StringIO.new(io)
        end
      end
      raise "Unable to create local socket" unless @io
    end
  end
end
```

and lib/fake_web/stub_socket.rb (add `close` method to socket stub)
```ruby
module FakeWeb
  class StubSocket
    def close
      @closed = true
    end
  end
end
```

mcfiredrill:
Closing in favor of https://github.com/vcr/vcr/pull/640

PR #639, opened 2017-02-21
Expect Faraday (with certain adapters at least) to support reason phrase

sbleon:
Most HTTP lib adapters in Faraday now expose the reason_phrase (AKA "status text") of a response. This is useful to me because some of the APIs that I work with put useful information, such as error messages, in the reason phrase. With this change, I can see the reason phrase in my recorded cassettes when using VCR with Faraday.


sbleon:
This replaces #634. I've rebased and attempted to get Travis to use a recent version of Faraday. We'll see how that goes!

sbleon:
This is not a breaking change. It should preserve compatibility with older versions of Faraday. It's just the tests that require a newer version with the new feature.

mcfiredrill:
Thanks for your PR!
The build looks good, I will try to take a closer look at this soon. 👍

sbleon:
@mcfiredrill could you take a look at this? It's a pretty small change to the library code. Thanks!

ecnalyr:
LGTM
@mcfiredrill Are there any reasons why we wouldn't want to merge this?

sbleon:
Thanks, Jason!

mcfiredrill:
Can't really see any reason not to merge. 👍

PR #645, opened 2017-04-21
add recompress_response option and hook

notEthan:
this adds a before_playback hook to recompress responses which VCR has decompressed on record. 

I love the option to decompress responses for readability and seeing changes in git diff, but the library I'm working with expects to get a gzipped response back, and errors if the response is not compressed.

I have not yet written tests. I wanted to gauge if this is a feature the maintainers would find desirable to include in VCR before I put in the time. if this would be merged, I'll do the tests.

I'm also not certain I'm not misusing the field `adapter_metadata` - it looked like the most appropriate place to put in information about whether / how VCR decompressed the response, but this isn't exactly related to the adapter.


krainboltgreene:
@notEthan Hey, we're interested! (A year later!) Lets get this updated from master and go forward.

notEthan:
I will try to find some time today or at least this week to get this up to date

PR #653, opened 2017-06-14
Add mutex to prevent multi-threading failures

jscattergood:
The mutex synchronizes the process of moving interactions from one list to another.  When being executed in multi-threaded code, two threads can see the state of the lists differently resulting in failures.

### DIFF
--- a/lib/vcr/cassette/http_interaction_list.rb
+++ b/lib/vcr/cassette/http_interaction_list.rb
@@ -22,21 +22,24 @@ module VCR
         @parent_list            = parent_list
         @used_interactions      = []
         @log_prefix             = log_prefix
+        @mutex                  = Mutex.new
 
         interaction_summaries = interactions.map { |i| "#{request_summary(i.request)} => #{response_summary(i.response)}" }
         log "Initialized HTTPInteractionList with request matchers #{request_matchers.inspect} and #{interactions.size} interaction(s): { #{interaction_summaries.join(', ')} }", 1
       end
 
       def response_for(request)
-        if index = matching_interaction_index_for(request)
-          interaction = @interactions.delete_at(index)
-          @used_interactions.unshift interaction
-          log "Found matching interaction for #{request_summary(request)} at index #{index}: #{response_summary(interaction.response)}", 1
-          interaction.response
-        elsif interaction = matching_used_interaction_for(request)
-          interaction.response
-        else
-          @parent_list.response_for(request)
+        @mutex.synchronize do
+          if index = matching_interaction_index_for(request)
+            interaction = @interactions.delete_at(index)
+            @used_interactions.unshift interaction
+            log "Found matching interaction for #{request_summary(request)} at index #{index}: #{response_summary(interaction.response)}", 1
+            interaction.response
+          elsif interaction = matching_used_interaction_for(request)
+            interaction.response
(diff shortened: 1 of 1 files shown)


jscattergood:
Build failures due to "An error occurred while installing rack (2.0.3), and Bundler cannot continue."

mcfiredrill:
@jscattergood 
Thanks for digging into that.
```
Gem::InstallError: rack requires Ruby version >= 2.2.2.
``` 
Can we install different versions of rack on different ruby versions?

saveriomiroddi:
@jscattergood @mcfiredrill @krainboltgreene I've tried the same approach (and jscattergood/vcr/master, for double checking), but it won't work.

The reason is that `HTTPInteractionList` is not the only instance hanging around. It's not immediately clear at a quick look why, but you can easily check by logging the `object_id`; on my project, you can see an output like this:

    >>> Performing HttpInteractionList#response_for from instance 21266040
    >>> Performing HttpInteractionList#response_for from instance 21266040
    >>> Performing HttpInteractionList#response_for from instance 20664680
    >>> Performing HttpInteractionList#response_for from instance 20664680
    >>> Performing HttpInteractionList#response_for from instance 20664680
    >>> Performing HttpInteractionList#response_for from instance 20664680
    >>> Performing HttpInteractionList#response_for from instance 20664680
    >>> Performing HttpInteractionList#response_for from instance 20664680
    >>> Performing HttpInteractionList#response_for from instance 20664680
    >>> Performing HttpInteractionList#response_for from instance 20664680

which shows that two different instances are being run, so the mutex won't yield the desired effect.

saveriomiroddi:
For reference, this PR, and the above comment, apply to #200.

krainboltgreene:
Is this still needed?

saveriomiroddi:
> Is this still needed?

This PR is a subset of the one I wrote (#690, which you've merged a short time ago), so it can be closed.

jscattergood:
@saveriomiroddi

You’ll have to track the lifecycle or that second interaction list.  There should be only one for the loaded cassette.  The cassette loading is usually done in a single thread, ie test thread.  You need to avoid initializing application threads until after you load the cassette.

My team has been use a monkey patched version of VCR with my fix for the last 8 months with no more issues.

I’m sorry but I don’t have time to get back into this right now, but I’m confident in my fix.

saveriomiroddi:
Hello!

I didn't intend that this PR doesn't work :smile: Just, that #690 is a superset of this, as it put two mutexes, one of which is the same of this PR (I came across it independently, as it's the immediately obvious place to look at).

The motivation for #690 is that a project of mine put extra threading pressure put on VCR, requiring not only `HttpInteractionList#response_for` to need a mutex, but `Cassette#http_interaction_list` too.

In the perspective of thread safety, I think no conclusion can be taken until somebody performs a serious review of the code - I don't doubt that even #690 is not sufficient for having true thread safety.

jscattergood:
Closing in favor of #690

krainboltgreene:
@jscattergood Sorry it took me so long to get to this, please feel free to add yourself to the credits list.

jscattergood:
@krainboltgreene That’s very gracious of you, but no worries.  

I’m very thankful for this project.  Keep up the good work!

PR #656, opened 2017-08-03
Fix on_body issue for Typhoeus w/ code style changes

ikapelyukhin:
#614 with the requested code style changes. Addresses #509.

PR #657, opened 2017-08-05
Fix Typhoeus/Faraday headers on playback

ikapelyukhin:
Apparently due to the way Faraday extracts headers from `Typhoeus::Request` on a live request the headers look like this:

```
 ["HTTP/1.1 200 OK ",
 "Content-Type: text/html;charset=utf-8",
 "Content-Length: 11",
 "Server: WEBrick/1.3.1 (Ruby/2.3.1/2016-04-26)",
 "Date: Sat, 05 Aug 2017 10:42:51 GMT",
 "Connection: Keep-Alive"]
```

When playing back the cassette, however, the HTTP version/status code line was missing, which broke Faraday:

```
["Content-Type: text/html;charset=utf-8",
 "Content-Length: 11",
 "Server: WEBrick/1.3.1 (Ruby/2.3.1/2016-04-26)",
 "Date: Sat, 05 Aug 2017 10:42:51 GMT",
 "Connection: Keep-Alive"]
```

### DIFF
--- a/lib/vcr/library_hooks/typhoeus.rb
+++ b/lib/vcr/library_hooks/typhoeus.rb
@@ -53,9 +53,10 @@ else
               :effective_url  => stubbed_response.adapter_metadata.fetch('effective_url', request.url),
               :mock           => true
 
+            first_header_line = "HTTP/#{stubbed_response.http_version} #{response.code} #{response.status_message}\r\n"
+            response.instance_variable_set(:@first_header_line, first_header_line)
             response.instance_variable_get(:@options)[:response_headers] =
-              "HTTP/#{response.http_version} #{response.code} #{response.status_message}\r\n" +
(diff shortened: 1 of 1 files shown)


isaacseymour:
Thanks for tracking this down! I've been seeing this with Faraday 0.12.2 and it's been very confusing! @mcfiredrill would be awesome to have this released soon 😄

ikapelyukhin:
Releasing a new version would be very good indeed. :grimacing:

ikapelyukhin:
OK, so the tests are passing now. Just re-ran them, didn't change anything.

ikapelyukhin:
@mcfiredrill OK, how about merging this PR? Then I could rebase #656, because that would fix 1 broken spec.

mcfiredrill:
Looks like it passed! Merging. 👍

PR #670, opened 2017-11-30
Output non-matched headers on error.

marshall-lee:
When matching on :headers it's quite difficult to understand why my http
calls don't match the cassettes.


marshall-lee:
Tried to fix the test in #671.

marshall-lee:
@krainboltgreene i rebased it off the `master` branch.

PR #691, opened 2018-02-04
Add .ruby-version to .gitignore

saveriomiroddi:
The `.ruby-version` file is the modern standard for Ruby versioning support (see RVM, rbenv, chruby...).

This PR adds it to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@
 
 # Ignore autoset environment variables
 /.ruby-env
+/.ruby-version
 
 # Ignore all bundler caching
 /vendor/cache


saveriomiroddi:
Does anybody know what's the `.ruby-env` for? It seems it's for versioning as well, but I couldn't find any reference on the internet about which software actually supports it. Its addition is dated 2015, I wonder if it's an obsolete standard.

krainboltgreene:
It's for `rbenv`, which might be a long forgotten thing now.

saveriomiroddi:
> It's for rbenv, which might be a long forgotten thing now.

ah, thanks. I was puzzled, because currently, rbenv uses `.ruby-version` and `.rbenv`, so I guess it was a past standard. should I remove that entry in this PR?

saveriomiroddi:
Hello! I've found that git supports local ignores, via `.git/info/exclude`, so this PR is not really required for devs working with multiple rubies/gemsets.

If you need me to do further work on in it in order to be merged, let me know! Otherwise, I'm fine if it's closed.

saveriomiroddi:
Closing for lack of activity.

PR #692, opened 2018-02-04
Remove the last Ruby 1.8 code

### LINKED ISSUE #676: cleanup old ruby 1.8 hacks
If you grep the source code for "1.8." you will probably still find a few hacks and things to support Ruby 1.8. Since we don't support 1.8 anymore these can be removed.

If anyone is looking for a place to start contributing this is a great way to start.

saveriomiroddi:
Remove Ruby 1.8 Ping code from util/internet_connection.rb.

This is the last Ruby 1.8 code, and closes #676.

### DIFF
--- a/lib/vcr/util/internet_connection.rb
+++ b/lib/vcr/util/internet_connection.rb
@@ -1,31 +1,25 @@
 module VCR
   # Ruby 1.8 provides Ping.pingecho, but it was removed in 1.9.
-  # So we try requiring it, and if that fails, define it ourselves.
-  begin
-    require 'ping'
-    Ping = ::Ping
-  rescue LoadError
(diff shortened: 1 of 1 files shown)


lib/vcr/util/internet_connection.rb
```diff
@@ -1,31 +1,25 @@
 module VCR
   # Ruby 1.8 provides Ping.pingecho, but it was removed in 1.9.
```
mathieujobin: why keeping this comment ?

saveriomiroddi: I think it's useful to provide background. With only the other comment (`This is copied, verbatim, from Ruby 1.8.7's ping.rb.`), it may not immediately obvious why this strategy is applied.

PR #697, opened 2018-03-06
Fix Code Climate badges URLs

epistrephein:
The Code Climate badges from shields.io are being refactored to comply with upstream changes, so the old URLs are not working anymore.
While we wait for the final version of the fix to be pushed to production, this PR adds support for the `maintainability` and `coverage` letter grade.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 vcr
 ===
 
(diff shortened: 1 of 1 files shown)


epistrephein:
Shields.io deploy for code climate is now live in prod, I've updated the coverage URL to the recommended one. This PR uses letter grade for maintainability and percentage grade for coverage, but both can be swapped now if you prefer different grading style.

PR #713, opened 2018-07-20
disable the webmock excon adapter

andrew-lewin:
Webmock changed when it gets enabled from when it's initially loaded to right before the rspec suite is ran. (see here for the diff https://github.com/bblimke/webmock/commit/d9af3a54d73c7de1cc16cf19e05b2522c4f7d093#diff-716564b83a1cd2a4dd19999e263d60f4)

This PR will make it so that if the webmock excon adapter is present, and we're using rspec, we disable the webmock excon adapter before the rspec test suite runs. This should keep vcr compatible with current versions of webmock


krainboltgreene:
Fantastic, thanks.

PR #716, opened 2018-08-02
We're going to switch to rubocop and circle ci

krainboltgreene:
- Remove other services
  - Setup rubocop
  - Setup circle ci

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,2 +1,0 @@
-ruby:
-  enabled: true


krainboltgreene:
@mcfiredrill I feel like circle really has upped their game in terms of how easy it is to setup complex workflows.

Also, while this sounds like dictation I am definitely looking for agreement from @vcr/core.

krainboltgreene:
@mcfiredrill Now that Travis-CI has decimated their staff, I feel even more strongly about this.

PR #730, opened 2018-12-18
Fix VCR.use_cassettes method implementation and spec

skryukov:
- update VCR.use_cassettes spec to test method arguments
- fix VCR.use_cassettes raises `no implicit conversion of nil into Hash`

### DIFF
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -202,10 +202,13 @@ module VCR
   #   # make multiple HTTP requests
   # end
   def use_cassettes(cassettes, &block)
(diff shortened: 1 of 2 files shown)


serggl:
@krainboltgreene can you please review this? 

while trying to use `use_cassettes` I also found 2 flaws:
1. `options` parameter is required, though it is not documented. if you try to use it without options provided it will crash:
```
VCR.use_cassettes([{ name: 'cassette a' }, { name: 'cassette b' }]) do {} =>

TypeError:
       no implicit conversion of nil into Hash
```

2. `options` parameter was ignored for the first cassette definition provided

I think this PR should fix both of these

Issue #737, opened 2019-03-05
CI tests broken "can't dup NilClass"


krainboltgreene:
Not much to go on, but apparently we broke 2.3 ruby.

PR #743, opened 2019-04-14
ruby 2.3 is EOL

gregmolnar:
Closes https://github.com/vcr/vcr/issues/737

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ before_install:
   - gem install bundler
 cache: bundler
 rvm:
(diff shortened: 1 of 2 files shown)


gregmolnar:
Looks like there are failures with newer versions of ruby too. I can reproduce them locally too. 
Will look into them as time permits.

gregmolnar:
It looks like `excon` `0.63.0` has some changes that brakes the spec suite. I locked the version to the last compatible one and the specs are passing now.

Issue #700, opened 2018-03-21
Add :single_cassette option to RSpec metadata


rachel-carvalho:
### Background

I've been testing an API call with VCR that I can't really re-do (it works a bit like a delete) because I _have_ to create it manually and use the resulting `id` in my spec. It's something like this:

```ruby
describe '#delete', :vcr do
  let(:foo) { API::Foo.new('ID that I can only get manually') }

  subject { foo.delete }

  it 'changes status' do
    expect { subject }.to change foo, :status
  end

  it 'does something else' do
    # ...
  end

  it 'does a third thing' do
    # ...
  end
end
```

In this situation, each of the `it` blocks records its own cassette. But I need a new `id` for each new request, so when I'm re-recording my cassettes, the first `it` passes and the subsequent ones fail.

I could just copy/paste the first cassette with matching names for the other blocks, but it doesn't feel right.

Then I found I could use [the `:cassette_name` option](https://relishapp.com/vcr/vcr/docs/test-frameworks/usage-with-rspec-metadata) which ensures all my `it`s use the same cassette.

So I changed my specs:

```ruby
describe '#delete' do
  let(:foo) { API::Foo.new('ID that I can only get manually') }

  subject { foo.delete }

  context 'success', vcr: { cassette_name: 'API_Foo/_delete/success' } do
    it 'changes status' do
      expect { subject }.to change foo, :status
    end

    # ...
  end

  context 'failure', vcr: { cassette_name: 'API_Foo/_delete/failure' } do
    # ...
  end
end
```

But now I'm manually writing the cassette names to keep the same naming scheme as the automatically generated cassettes.

### Proposed solution

It would be really cool if we could specify we want a single cassette to be generated based on the block path, like this:

```ruby
describe '#delete' do
  subject { foo.delete }

  context 'success', vcr: { single_cassette: true } do
    # ...
  end

  context 'failure', vcr: { single_cassette: true } do
    # ...
  end
end
```

Sorry for the long wall of text and code, I couldn't find a succinct way to explain this.

I'd love have a go at this if you guys think it's a good idea (although I'll be leaving on vacation soon and will probably not have access to a computer for at least a month, but I'll do it then).

I realize there might be something I'm not doing right here so I'm open to suggestions that don't include adding a new feature :)

mcmire:
I just came across into this and I'm curious if I want a similar feature than you. Would `single_cassette` generate a cassette for the entire context as a whole instead of each test individually?

[...]

PR #761, opened 2019-07-17
Add single cassette

dieggofaustino:
The `single_cassette` option makes every test case below the context/describe with this option use the description of this block as cassette name. It has the same behaviour as using cassette_name option in the context/describe, but automatically generating the name.

This PR solves the issue https://github.com/vcr/vcr/issues/700


ecnalyr:
I like this, but I didn't _get it_ until after I read the related issue a couple of times.

Could you add something to the documentation that explains what this `single_cassette` option does?

PR #764, opened 2019-07-31
Fix turning off VCR not working when server thread already started

cgunther:
## Background

When VCR is first used in the main (test) thread, it initializes a context, holding the status of whether VCR is turned off. When another thread is started (ie. app server), VCR initializes a context there by duplicating the main thread's context and linking to the main thread's context.

If you turned off VCR in the main thread after it's context was duplicated to an additional thread, the additional thread wouldn't see the new turned off status (since it was only on the linked context), so it'd fail thinking VCR was still turned on but no cassette was inserted when a request was made.

## Solution

Now we check both the current thread's context, and the linked (main) thread's context to see if VCR is turned off, to ensure an app server thread can see when VCR is turned off on the main thread.

## Details

This was most noticeable in Capybara tests with a JS driver where the app server thread was started once upon the first spec needing it, thus duplicating the main thread's context then, but if later specs turned off VCR, the app server thread never had it's context updated to reflect that, nor did it check the main thread's context to see if VCR was turned off there.

### DIFF
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -308,6 +308,9 @@ module VCR
   # @see #turn_off!
   # @see #turned_off
   def turned_on?
+    linked_context = current_context[:linked_context]
+    return !linked_context[:turned_off] if linked_context
+
     !context_value(:turned_off)
   end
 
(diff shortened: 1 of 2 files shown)


yiyuanc:
this looks good to me!

olleolleolle:
(I added formatting markup to the Description of this PR, to ease navigation and skim-reading.)

cgunther:
Thanks for the merge!

PR #765, opened 2019-08-07
Add record_on_error configuration option

martijnversluis:
When you are TDD'ing an API integration and an error occurs it will
often be caused by an bug in the implementation. In that case no
interactions should be recorded, to avoid having to manually throw away
the cassette every test run until all bugs are solved.

When setting `record_on_error` to false, no interactions will be written
to disk when an error is raised inside the block that uses the cassette.

This change is inspired by [Walkman](https://github.com/derekkraan/walkman), which by default does not record a tape when your test fails.

### DIFF
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -186,6 +186,9 @@ module VCR
 
     begin
       call_block(block, cassette)
+    rescue StandardError
+      cassette.run_failed!
+      raise
     ensure
       eject_cassette
     end
--- a/features/configuration/debug_logging.feature
+++ b/features/configuration/debug_logging.feature
@@ -34,7 +34,7 @@ Feature: Debug Logging
     Given that port numbers in "record.log" are normalized to "7777"
     Then the file "record.log" should contain exactly:
       """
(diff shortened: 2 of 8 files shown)


krainboltgreene:
What a fantastic pull request.

lib/vcr.rb
```diff
@@ -186,6 +186,9 @@ def use_cassette(name, options = {}, &block)
 
     begin
       call_block(block, cassette)
+    rescue StandardError
+      cassette.run_failed!
+      raise
```
krainboltgreene: We should raise the exception that gets caught or an exception that wraps the exception caught.

martijnversluis: `raise` without arguments should re-raise the last exception that was caught (https://ruby-doc.org/core-2.1.2/Kernel.html#method-i-raise). I can agree it would improve readability to explicitly rescue and re-raise the error. Or is there a functional disadvantage to using `raise` with no arguments?

krainboltgreene: :shocked_pikachu:

PR #769, opened 2019-09-14
Fix #746: Add drop_unused_requests option to exclude unused interactions

### LINKED ISSUE #746: New record mode "re_record"
At first, thanks for the VCR gem! We're using it quite frequently and it's doing a great job.

**Is your feature request related to a problem? Please describe.**
We're using VCR to stub slow/unstable HTTP communication in tests. Sometimes changed code provokes differend requests, thusly requiring a re-recording of cassettes.

The amount of VCR cassettes (>200) plus the separation of test from cassette location makes re-recording a little tedious. Say I need to re-record all search tests, I need to look and find where the concerned cassettes are stored, `rm` them manually, and then run all tests that lost their cassette (only those, at best).

**Describe the solution you'd like**
If VCR had a (default cassette option) `record_mode: :re_record`, I would turn it on and just run the tests that I want to drop and re-record their cassettes. Afterwards I would switch back to `:once` mode. The idea is that the developer can re-record just by running tests.

(Probably I'd then funnel the activation of the `re_record` mode through an ENV var that's read in the VCR config, but that's out of scope. Example: `VCR=re_record <run tests>`)

**Describe alternatives you've considered**
While the automatic re-recording feature sounds handy, we cannot afford to have tests "randomly" drop their cassettes. It slows down the test suite run at unexpected times, and if requests fail, the developer faces broken tests he didn't break.

The `record_mode: :all` is neither a solution. While it updates previous requests and adds new requests, it does not drop unused requests. Together with the `allow_unused_http_interactions: false` option, this is not a feasible solution.

What do you think?

mhuntglickman:
Address the ability to drop unused requests in cassettes. 

This seemed a better way to address the needs of all vs a whole new recording method.

Fixes #746

### DIFF
--- a/lib/vcr/cassette.rb
+++ b/lib/vcr/cassette.rb
@@ -176,7 +176,7 @@ module VCR
         :record, :record_on_error, :erb, :match_requests_on, :re_record_interval, :tag, :tags,
         :update_content_length_header, :allow_playback_repeats, :allow_unused_http_interactions,
         :exclusive, :serialize_with, :preserve_exact_body_bytes, :decode_compressed_response,
-        :recompress_response, :persist_with, :clean_outdated_http_interactions
+        :recompress_response, :persist_with, :clean_outdated_http_interactions, :drop_unused_requests
       ]
 
       if invalid_options.size > 0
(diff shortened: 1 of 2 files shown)


thezainsaleem:
@mhuntglickman Please Back Merge

ryanseys:
I rebased this off the latest master in this new PR: https://github.com/vcr/vcr/pull/946

ryanseys:
https://github.com/vcr/vcr/pull/946 is now merged. Not yet released, but we can close this.

olleolleolle:
(Closed in favor of #946.)

PR #772, opened 2019-09-23
Update vcr.gemspec

luke-hill:
Allow aruba to use latest version in stable 0.14 series

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ platform :mri do
   gem "typhoeus", "~> 1.1.0"
   gem "patron", "0.6.3"
   gem "em-http-request"
-  gem "curb", "~> 0.8.8"
+  gem "curb", "~> 0.9.0"
 end
 
 platform :ruby do
--- a/Gemfile
+++ b/Gemfile
@@ -8,7 +8,7 @@ gem 'jruby-openssl', :platforms => :jruby
 
 platform :mri do
   gem "typhoeus", "~> 1.1.0"
-  gem "patron", "0.6.3"
+  # gem "patron", "0.6.3"
   gem "em-http-request"
   gem "curb", "~> 0.9.0"
 end
--- a/features/step_definitions/cli_steps.rb
+++ b/features/step_definitions/cli_steps.rb
@@ -65,7 +65,7 @@ module VCRHelpers
   end
 
   def modify_file(file_name, orig_text, new_text)
-    in_current_dir do
+    cd('.') do
       file = File.read(file_name)
       regex = /#{Regexp.escape(orig_text)}/
       expect(file).to match(regex)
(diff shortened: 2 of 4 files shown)


luke-hill:
If any of the current vcr maintainers want to take a look at this feel free.

I was going through reverse dependencies of aruba and found this is massively behind. There is likely some refactoring work needed to be done, a similar PR has been raised and fixed for most of the rspec implementations.

olleolleolle:
Thanks for looking at reverse dependencies, neighbor!

Using latest stable Aruba is reasonable, and seems uncontroversial to me. It's being maintained and cared for by a team which is active. 

I hope that we can get to investigate the CI failures, and if we have detailed questions, about changes/history-of-aruba, we'll ask them in this thread.

olleolleolle:
The warnings output is very helpful and detailed. 

Example at https://travis-ci.org/vcr/vcr/jobs/588334309

soulcutter:
Potentially controversial opinion... Aruba is pre-1.0, so per SemVer any release could result in breaking changes. Should the version pin be to `>=` or `>` rather than `~>`? I could see that aggressively breaking things downstream sometimes, but hopefully it would result in keeping this more up-to-date (or hey, maybe Aruba hits a major version)

soulcutter:
Having slept on it, seeing where Aruba is version-wise, and digesting the reply... I retract my pinning suggestion. The scope of this change seems right :+1: apologies for the distraction!

luke-hill:
Not a problem, now the next issue is getting it working. We want all major reverse dependencies to be able to upgrade to 1.0 when the time is right. And currently a healthy amount of tests here fail. So lets work out if it's bugs in test code or aruba

luke-hill:
I've made some slight tweaks to my changes and ran things locally (Which in itself caused a huge plethora of issues)

RSpec (with patron commented out of the bundle and curb slightly bumped)
```
Finished in 18.83 seconds (files took 1.66 seconds to load)
1944 examples, 216 failures
```

Cucumber - hit a runtime error for an undefined or ambiguous cuke
```
aruba-0.8.1/lib/aruba/cucumber.rb:68:in `/^(?:a|the) directory(?: named)? "([^"]*)" does not exist$/'
features/step_definitions/cli_steps.rb:84:in `/^the directory "([^"]*)" does not exist$/'
```

luke-hill:
@soulcutter if you could have a look and maybe try help with where I've got to so far that'd be good.

I've opted to go for an 0.8.x version of aruba first. Baby steps. This is one other repos have used and involved much less fixes and warnings being thrown.

If we can get this upgrade working I propose we then target the 0.14.x revision afterwards.

luke-hill:
@soulcutter or any other vcr maintainer.

I've had a stab at upgrading to 0.8.0 from your current version and encountered some trickiness. But the main issue was I cannot get the physical suite to run reliably.

Could one of you either take over this PR, or have an attempt yourself at upgrading to 0.8.0 - That would be a reasonable upgrade chunk, plus it would fix up some of the issues that you may encounter later down the line.

We're trying to get all the big gems upgraded to a `0.14.x` version so when we upgrade to v1 they're not too shocked. At the moment you're quite far back so we need to try help you.

krainboltgreene:
The failures in this pull request seem very specific. I'll look into it tonight.

luke-hill:
Cool stuff. I will try help however I can. It's likely that things will need altering here (And maybe in aruba too).

The only thing to note is that any fixes to aruba would need to be done in the `0.14.x` series. So we may need to patch them here, update to 0.14 here and then remove the patches and fix in aruba.

I'm aware that sounds convoluted. But we're trying to slowly get aruba fixed up so when we do update to 1.0 every major consumer isn't left behind.

olleolleolle:
@luke-hill It seems @mvz got many of these changes into `master` via the just-merged PR.

Did that cover all of it? Can this PR now be closed?

Thanks for offering all this work for vcr. IT'S AWESOME.

luke-hill:
These commits are already in master as he cp-ed them

PR #782, opened 2019-12-19
Carry "Remove trailing whitespaces from YAML files"

olleolleolle:
This PR is a rebase of the two commits in the PR #591. 

Thanks for offering it, @exAspArk. 

Closes #591

### DIFF
--- a/lib/vcr/cassette/serializers/psych.rb
+++ b/lib/vcr/cassette/serializers/psych.rb
@@ -28,7 +28,9 @@ module VCR
         # @return [String] the YAML string
         def serialize(hash)
           handle_encoding_errors do
-            ::Psych.dump(hash)
+            result = ::Psych.dump(hash)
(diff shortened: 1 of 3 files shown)


lib/vcr/cassette/serializers/psych.rb
```diff
@@ -28,7 +28,9 @@ def file_extension
         # @return [String] the YAML string
         def serialize(hash)
           handle_encoding_errors do
-            ::Psych.dump(hash)
+            result = ::Psych.dump(hash)
+            result.gsub!(": \n", ": null\n") # set canonical null value in order to avoid trailing whitespaces
+            result
```
krainboltgreene:
```suggestion
            ::Psych.dump(hash).gsub(": \n", ": null\n") # set canonical null value in order to avoid trailing whitespaces
```

olleolleolle: My guess: the in-place substitution, which doesn't create new string value, is cheaper. Do we care, here?

PR #787, opened 2020-01-08
CI: Add 2.7 to the matrix

olleolleolle:
This PR adds the Ruby 2.7 element to the matrix.

Current problem: We run 2.0.2 of cucumber - `spec.add_development_dependency "cucumber", "~> 2.0.2"` which has trouble with 2.7's Proc.


olleolleolle:
The failing build is with Faraday 1.0, which has a different order to its middlewares.

olleolleolle:
>       2 /home/travis/build/vcr/vcr/lib/vcr/structs.rb:459: warning: Using the last argument as keyword parameters is deprecated

This needs to be fixed.

olleolleolle:
Current thing to learn: Why does not the 2.7 in Travis CI install em-http-request? It installs on my Mac, fine, when I use 2.7.

krainboltgreene:
Wait, does this mean that 2.7 broke expectations? That's a bit nuts.

olleolleolle:
@krainboltgreene The problem w/ Cucumber 2.0.2 was that it checked something re: `Proc.new` (which is gone in 2.7) - IIRC

krainboltgreene:
Thoughts on going forward? 2.7 coverage would be fantastic.

olleolleolle:
Since we have [a long version shadow: `">= 1.9.3"`.](https://github.com/vcr/vcr/blob/master/vcr.gemspec#L22)

Here's one idea: in case we _can't_ update away from Cucumber 2.0.2.

- a [testing seam](https://www.informit.com/articles/article.aspx?p=359417&seqNum=2) using a kind of a preprocessor seam: Let's imagine a `spec/overrides/` directory, where can place a hacked version of the cucumber 2.0.2 file which supports the 2.7 syntax

olleolleolle:
This quote is from the Versions section of the `README`:

> Note that as of VCR 3, 1.8.7 and 1.9.2 are explicitly not supported.

The `vcr.gemspec` file has 1.9.3+ as required Ruby version. So, there's been "ditch versions" decisions in the past.

To support 2.7 and the future, we need to upgrade the lowest level to something newer. What should that "newer" be?

Ruby 2.5 is in-support, on "security fixes only", which it will be for < 1 year more.

[Ruby Maintenance Branches](https://www.ruby-lang.org/en/downloads/branches/)

([Current cucumber's gemspec](https://github.com/cucumber/cucumber-ruby/blob/024caa709ca173a0d904c17a60428bb1e3e8b5c9/cucumber.gemspec) says `s.required_ruby_version = '>= 2.3'`, so that's a sort of _minimum_.)

PR #790, opened 2020-01-22
Add Ruby 2.7 to Travis-CI matrix

bastelfreak:
Ruby 2.7 got released some time ago. The first linux distributions
already started to ship. Hence I think it makes sense to also run the
tests on this version.


krainboltgreene:
See #787

PR #800, opened 2020-02-20
Test against 2.7

BrianHawley:


BrianHawley:
I've been running into a bug in Ruby 2.7 in code from this gem and I'm trying to see if it can be reproduced here.

BrianHawley:
This test failure is not the one I was looking for. It still seems worth fixing though.

Issue #652, opened 2017-06-12
Metadata Case Issues


lackita:
We're running into some trouble with some of our team on linux and some on mac. The problem is that the directories generated when using rspec metadata are lowercase when run on mac, but camel case when run on linux. This seems to be an issue with mac's filesystem being case-insensitive, has anybody found a good work-around for the problem?

PR #802, opened 2020-02-28
Add option to downcase cassette names before saving

mctaylorpants:
Hello! 👋 I'd like to propose an option to the default `FileSystem` persister that allows cassette names to be downcased before saving and retrieving, in order to solve case-sensitivity issues between filesystems.

### Background
At Clio, our developers use macOS (which, by default, has a case _insensitive_ filesystem), but we run CI on Linux. We use RSpec, and usually allow VCR to generate cassette names based on the spec descriptions.

We recently had an issue where a cassette had already been generated for a test, but the test's description was modified slightly (i.e. `it "makes an http request"` became `it "makes an HTTP request"`). This passed fine locally, because macOS didn't care about the difference in case, but failed on CI.

### The implementation
I've added a `persister_options` hash to allow custom options to be passed to the persister. I've added an option, `downcase_cassette_names`, to support the feature. (Let me know if you think there could be a better name for this.)

### Usage example
Pass additional configuration to VCR:

```ruby
VCR.configuration do |config|
  ...
    config.default_cassette_options = { 
      persister_options: { downcase_cassette_names: true }
    }
  ...
end
```

### Impact
Since this must be explicitly enabled, it will not cause any breaking changes. However, if it's enabled and users have pre-existing cassettes with uppercase letters, specs may fail on Linux. If users decide to enable this option, they'll have to rename existing cassettes to downcase the names. (this is easier said than done on macOS, so [I wrote a script](https://gist.github.com/mctaylorpants/04ceca3d4d0f7230c191d39e6462b4e5) that gets the job done).

Fixes https://github.com/vcr/vcr/issues/652

### DIFF
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -108,7 +108,11 @@ module VCR
   #  use. Defaults to :file_system. You can also register and use a
   #  custom persister.
   # @option options :persister_options [Hash] Pass options to the
-  # persister specified in `persist_with`.
+  #  persister specified in `persist_with`. Currently available options for the file_system persister:
+  #    - `:downcase_cassette_names`: when `true`, names of cassettes will be
+  #      normalized in lowercase before reading and writing, which can avoid
+  #      confusion when using both case-sensitive and case-insensitive file
+  #      systems.
   # @option options :preserve_exact_body_bytes [Boolean] Whether or not
   #  to base64 encode the bytes of the requests and responses for this cassette
   #  when serializing it. See also `VCR::Configuration#preserve_exact_body_bytes`.
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -107,6 +107,8 @@ module VCR
   # @option options :persist_with [Symbol] Which cassette persister to
   #  use. Defaults to :file_system. You can also register and use a
   #  custom persister.
+  # @option options :persister_options [Hash] Pass options to the
+  # persister specified in `persist_with`.
   # @option options :preserve_exact_body_bytes [Boolean] Whether or not
   #  to base64 encode the bytes of the requests and responses for this cassette
   #  when serializing it. See also `VCR::Configuration#preserve_exact_body_bytes`.
--- a/lib/vcr/cassette/persisters/file_system.rb
+++ b/lib/vcr/cassette/persisters/file_system.rb
@@ -61,8 +61,9 @@ module VCR
         end
 
         def downcase_cassette_names?
-          options = VCR.configuration.default_cassette_options[:persister_options]
-          options && !!options[:downcase_cassette_names]
+          !!VCR.configuration
+            .default_cassette_options
+            .dig(:persister_options, :downcase_cassette_names)
         end
       end
     end
--- a/lib/vcr/cassette/persisters/file_system.rb
+++ b/lib/vcr/cassette/persisters/file_system.rb
@@ -55,7 +55,14 @@ module VCR
             file_extension = '.' + parts.pop
           end
 
-          parts.join('.').gsub(/[^[:word:]\-\/]+/, '_') + file_extension.to_s
+          file_name = parts.join('.').gsub(/[^[:word:]\-\/]+/, '_') + file_extension.to_s
+          file_name.downcase! if downcase_cassette_names?
+          file_name
+        end
+
+        def downcase_cassette_names?
+          options = VCR.configuration.default_cassette_options[:persister_options]
+          options && !!options[:downcase_cassette_names]
         end
       end
     end
--- a/lib/vcr/configuration.rb
+++ b/lib/vcr/configuration.rb
@@ -496,7 +496,8 @@ module VCR
         :match_requests_on => RequestMatcherRegistry::DEFAULT_MATCHERS,
         :allow_unused_http_interactions => true,
         :serialize_with    => :yaml,
-        :persist_with      => :file_system
+        :persist_with      => :file_system,
+        :persister_options => {}
       }
 
       self.uri_parser = URI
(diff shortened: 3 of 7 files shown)


chlorophyll95:
Just wondering if this change should even be optional or just enforced for the names of all cassettes?

spec/lib/vcr/cassette/persisters/file_system_spec.rb
```diff
@@ -64,12 +64,23 @@ class Persisters
...
           end
 
           it 'handles files with no extensions (even when there is a dot in the path)' do
             expected = File.join(FileSystem.storage_location, "/foo_bar/baz_qux")
             expect(FileSystem.absolute_path_to_file("/foo.bar/baz qux")).to eq(expected)
           end
+
+          it 'downcases cassette names if the option is passed' do
+            VCR.configuration.default_cassette_options.merge!(
+              { :persister_options => { :downcase_cassette_names => true }})
```
bstaudacher: The parenth linting here is bugging me out, would expect the closing one to be on a newline.

lib/vcr/configuration.rb
```diff
@@ -496,7 +496,8 @@ def initialize
         :match_requests_on => RequestMatcherRegistry::DEFAULT_MATCHERS,
         :allow_unused_http_interactions => true,
         :serialize_with    => :yaml,
-        :persist_with      => :file_system
+        :persist_with      => :file_system,
+        :persister_options => {}
```
bstaudacher: What is the reasoning for introducing a full set of options here? I would have expected something more along the lines of a single option ```:downcase_cassette_names => false```.

mctaylorpants:
> Just wondering if this change should even be optional or just enforced for the names of all cassettes moving forward?

@chlorophyll95 good question - it would be nice if we could enforce it, but I believe it would break existing cassettes if they have uppercase characters. `#sanitized_file_name_from` can get called from a few places in the lifecycle of a cassette: once when VCR is about to write a new cassette, and once when VCR is determining if a cassette already exists for the given spec. It's this last point that will cause existing cassettes to break. For that reason, it may be best if we keep it optional for now (or make it on by default and bundle it with a major version upgrade).

lib/vcr/configuration.rb
mctaylorpants: I was following a pattern here that I've seen often in Rails, where a given configuration option will have its own option hash if it's destined to be used by that other piece of config. For example, you can set `config.session_store` to `:memory`, `:redis`, etc, and whichever adapter you specify there can be further customized by what you pass to `config.session_options`.

I do agree it might be a bit overkill since there's only one supported option right now, so I'm happy to change this if others feel the same.

lib/vcr/cassette/persisters/file_system.rb
```diff
@@ -55,7 +55,14 @@ def sanitized_file_name_from(file_name)
...
           end
 
-          parts.join('.').gsub(/[^[:word:]\-\/]+/, '_') + file_extension.to_s
+          file_name = parts.join('.').gsub(/[^[:word:]\-\/]+/, '_') + file_extension.to_s
+          file_name.downcase! if downcase_cassette_names?
+          file_name
+        end
+
+        def downcase_cassette_names?
+          options = VCR.configuration.default_cassette_options[:persister_options]
```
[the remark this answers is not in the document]
mctaylorpants: Love it! Done.

lib/vcr.rb
```diff
@@ -107,6 +107,8 @@ def current_cassette
   # @option options :persist_with [Symbol] Which cassette persister to
   #  use. Defaults to :file_system. You can also register and use a
   #  custom persister.
+  # @option options :persister_options [Hash] Pass options to the
+  # persister specified in `persist_with`.
```
olleolleolle: If a reader were here - would they know anywhere in the docs that the option available to them was "downcase..."? Can we put it in here, to make it discoverable within the rendered API docs?

olleolleolle: (This is not a blocker to merging, I hasten to say.)

mctaylorpants: How about something like this? (I'm not too familiar with how much formatting these API docs support so let me know if there's a better way:)

```suggestion
  #  persister specified in `persist_with`. Currently available options:
  #    - downcase_cassette_names: when `true`, names of cassettes will be
  #      normalized in lowercase before reading and writing, which can avoid
  #      confusion when using both case-sensitive and case-insensitive file
  #      systems.
```

olleolleolle: In order to find out if your edit looks good, you can run `yard`, the documentation tool, and then open doc/index.html, navigate to the rendered HTML, and see the rendering.

olleolleolle: The language is 👍 💯 though!

mctaylorpants: Cool! I'll give `yard` a shot on Monday, I don't have access to my dev machine at the moment.

olleolleolle: I'll render a preview here, as a picture:

image

olleolleolle: (This was after some minor fiddling. 

I was using the `yard server --reload` mode. Re-renders the viewed page on reload in the browser.)

mctaylorpants: Looks good, thanks for the screenshot! I updated my original suggestion with your edits. I also added clarification that the option is for the file_system persister.

mctaylorpants:
Hey @olleolleolle, @krainboltgreene, just wondering what the status of this PR is - are we OK to merge?

zdennis:
It doesn't look like this has been released yet? Are there any plans to do a VCR release in the near future?

lib/vcr/cassette/persisters/file_system.rb
```diff
@@ -55,7 +55,15 @@ def sanitized_file_name_from(file_name)
...
-          parts.join('.').gsub(/[^[:word:]\-\/]+/, '_') + file_extension.to_s
+          file_name = parts.join('.').gsub(/[^[:word:]\-\/]+/, '_') + file_extension.to_s
+          file_name.downcase! if downcase_cassette_names?
+          file_name
+        end
+
+        def downcase_cassette_names?
+          !!VCR.configuration
+            .default_cassette_options
+            .dig(:persister_options, :downcase_cassette_names)
```
zdennis, line 66: Should the RSpec and other integrations be updated to support this option? 

I'm looking at https://github.com/vcr/vcr/blob/master/lib/vcr/test_frameworks/rspec.rb#L47-L48

PR #809, opened 2020-03-30
Fix tests with excon >= 0.63

### LINKED ISSUE #801: Fix test failures with excon 0.72
We'd like to update ruby-excon to 0.72 in debian but 4 tests are failing with error messages like this

```
  1) Excon hook behaves like a hook into an HTTP library using Excon .stub_requests using specific match_attributes :headers returns the expected response for a {"X-Http-Header1"=>"val1"} request
     Failure/Error: raise VCR::Errors::UnhandledHTTPRequestError.new(vcr_request)
     
     VCR::Errors::UnhandledHTTPRequestError:
     
     
       ================================================================================
       An HTTP request has been made that VCR does not know how to handle:
         GET http://wrong-domain.com/wrong/path
     
       There is currently no cassette in use. There are a few ways
       you can configure VCR to handle this request:
     
         * If you're surprised VCR is raising this error
           and want insight about how VCR attempted to handle the request,
           you can use the debug_logger configuration option to log more details [1].
         * If you want VCR to record this request and play it back during future test
           runs, you should wrap your test (or this portion of your test) in a
           `VCR.use_cassette` block [2].
         * If you only want VCR to handle requests made while a cassette is in use,
           configure `allow_http_connections_when_no_cassette = true`. VCR will
           ignore this request since it is made when there is no cassette [3].
         * If you want VCR to ignore this request (and others like it), you can
           set an `ignore_request` callback [4].
     
       [1] https://www.relishapp.com/vcr/vcr/v/5-1-0/docs/configuration/debug-logging
       [2] https://www.relishapp.com/vcr/vcr/v/5-1-0/docs/getting-started
       [3] https://www.relishapp.com/vcr/vcr/v/5-1-0/docs/configuration/allow-http-connections-when-no-cassette
       [4] https://www.relishapp.com/vcr/vcr/v/5-1-0/docs/configuration/ignore-request
       ================================================================================
     Shared Example Group: "a hook into an HTTP library" called from ./spec/lib/vcr/library_hooks/excon_spec.rb:15
     # ./lib/vcr/request_handler.rb:97:in `on_unhandled_request'
     # ./lib/vcr/request_handler.rb:24:in `handle'
     # ./lib/vcr/middleware/excon.rb:68:in `before_request'
     # ./lib/vcr/middleware/excon.rb:20:in `request_call'
     # ./vendor/bundle/ruby/2.5.0/gems/excon-0.72.0/lib/excon/middlewares/mock.rb:57:in `request_call'
     # ./vendor/bundle/ruby/2.5.0/gems/excon-0.72.0/lib/excon/middlewares/instrumentor.rb:34:in `request_call'
     # ./vendor/bundle/ruby/2.5.0/gems/excon-0.72.0/lib/excon/middlewares/idempotent.rb:19:in `request_call'
     # ./vendor/bundle/ruby/2.5.0/gems/excon-0.72.0/lib/excon/middlewares/base.rb:22:in `request_call'
     # ./vendor/bundle/ruby/2.5.0/gems/excon-0.72.0/lib/excon/middlewares/base.rb:22:in `request_call'
     # ./vendor/bundle/ruby/2.5.0/gems/excon-0.72.0/lib/excon/middlewares/base.rb:22:in `request_call'
     # ./vendor/bundle/ruby/2.5.0/gems/excon-0.72.0/lib/excon/connection.rb:275:in `request'
     # ./spec/support/http_library_adapters.rb:226:in `make_http_request'
     # ./spec/support/shared_example_groups/hook_into_http_library.rb:471:in `make_http_request'
     # ./spec/support/shared_example_groups/hook_into_http_library.rb:429:in `block (3 levels) in matching_on'
```

Updated excon version to 0.72.0 in gemspec and ran bundle install, bundle exec rake

Ruby 2.5
Gem 5.1.0
HTTP same as the version that passed
Mock same as the version that passed
Rails {{Replace with version if applicable}}
Rspec {{Replace with version if applicable}}

dleidert:
In this version an 'Accept' header has been added in excon [1] and needs to be
added to the default headers to merge too. This should fix #801 and with the
version check being backwards compatible

[1] https://github.com/excon/excon/blame/master/lib/excon/constants.rb#L141

### DIFF
--- a/spec/support/http_library_adapters.rb
+++ b/spec/support/http_library_adapters.rb
@@ -227,7 +227,11 @@ HTTP_LIBRARY_ADAPTERS['excon'] = Module.new do
   end
 
   def normalize_request_headers(headers)
-    headers.merge('User-Agent' => [Excon::USER_AGENT])
+    if Excon::VERSION.to_f < 0.63
+      headers.merge('User-Agent' => [Excon::USER_AGENT])
+    else
+      headers.merge('User-Agent' => [Excon::USER_AGENT], 'Accept' => ['*/*'])
+    end
   end
 end
 
--- a/spec/support/http_library_adapters.rb
+++ b/spec/support/http_library_adapters.rb
@@ -227,7 +227,7 @@ HTTP_LIBRARY_ADAPTERS['excon'] = Module.new do
   end
 
   def normalize_request_headers(headers)
-    if Excon::VERSION.start_with?("0.63")
+    if Gem::Version.new(Excon::VERSION) < Gem::Version.new("0.63")
       headers.merge('User-Agent' => [Excon::USER_AGENT])
     else
       headers.merge('User-Agent' => [Excon::USER_AGENT], 'Accept' => ['*/*'])
--- a/spec/support/http_library_adapters.rb
+++ b/spec/support/http_library_adapters.rb
@@ -227,7 +227,7 @@ HTTP_LIBRARY_ADAPTERS['excon'] = Module.new do
   end
 
   def normalize_request_headers(headers)
-    if Excon::VERSION.to_f < 0.63
+    if Excon::VERSION.start_with?("0.63")
       headers.merge('User-Agent' => [Excon::USER_AGENT])
     else
       headers.merge('User-Agent' => [Excon::USER_AGENT], 'Accept' => ['*/*'])


spec/support/http_library_adapters.rb
```diff
@@ -227,7 +227,11 @@ def make_http_request(method, url, body = nil, headers = {})
   end
 
   def normalize_request_headers(headers)
-    headers.merge('User-Agent' => [Excon::USER_AGENT])
+    if Excon::VERSION.to_f < 0.63
```
krainboltgreene:
```suggestion
    if Excon::VERSION.start_with?("0.63")
```

So two reasons I want this change:

  1. Not all semantic versions can be floats
  2. Since Excon is sadly still unstable (`0.X.Y`) we can't safely say *any* major/minor above `0.63.X` is safe.

olleolleolle: Could we compare Strings?

```
2.6.3 :001 > '0.6.0' < '0.7.3'
 => true
```

dleidert: I see your point and it seems to work fine.

JFTR: I assume that it is safe for all versions higher then 0.63 because the change will unlikely be reverted (if you read the commit, why they introduced this header). I've tested locally with excon 0.60 and excon 0.72.

dleidert:
Hm. The travis build failed with the same errors. I'll check, if there Accept header has been already added in 0.62.

spec/support/http_library_adapters.rb
terceiro: strictly speaking the correct way of doing this tests is 

```ruby
Gem::Version.new(Excon::VERSION) < Gem::Version.new("0.63")
```

krainboltgreene: I had actually forgotten that Gemspec comes with a version comparison API. I had assumed it didn't.

krainboltgreene:
@dleidert Thanks for the work on this, it's much appreciated.

dleidert:
Hm. Nope, excon 0.62 does not contain this header. I have to wonder, what's happening here.

spec/support/http_library_adapters.rb
```diff
@@ -227,7 +227,11 @@ def make_http_request(method, url, body = nil, headers = {})
   end
 
   def normalize_request_headers(headers)
-    headers.merge('User-Agent' => [Excon::USER_AGENT])
+    if Excon::VERSION.start_with?("0.63")
```
olleolleolle:
```suggestion
    if Gem::Version.new(Excon::VERSION) < Gem::Version.new("0.63")
```

Thanks @terceiro for creating this thingie, perhaps it could help here?

dleidert: I'm testing this too. As soon as I have the travis results I'll get back to you.

dleidert: Argh. The change in <https://github.com/vcr/vcr/pull/809/commits/da0714182b260022b7bb185ce8bac53d827134c5> changed the test logic and therefor the Accept header got added for versions lower 0.63. @terceiro fixed the test logic again.

dleidert:
Looks good.

Issue #806, opened 2020-03-22
Possibly confusing "MIT-Hippocratic-1.2" license "tag" in vcr.gemspec


pombredanne:
The current vcr.gemspec has this license declared: `spec.license       = "MIT-Hippocratic-1.2"`
https://github.com/vcr/vcr/blob/1f70077f73d176ab577a50033b6762b7a6d70772/vcr.gemspec#L16
I am not sure how this relate to the licensing of the package as from what I can see the license is the Hippocratic 2.0 (and there are remaining MIT obligations), but `MIT-Hippocratic-1.2` --if it means anything-- does not seem to be neither current nor correct

krainboltgreene:
Great catch, thanks!

pombredanne:
See https://github.com/vcr/vcr/pull/810#pullrequestreview-384078022  
I think there may still be some issues with the MIT part?

PR #810, opened 2020-03-30
Correct license slug

krainboltgreene:
#806

### DIFF
--- a/vcr.gemspec
+++ b/vcr.gemspec
@@ -13,7 +13,7 @@ Gem::Specification.new do |spec|
   spec.summary       = %q{Record your test suite's HTTP interactions and replay them during future test runs for fast, deterministic, accurate tests.}
   spec.description   = spec.summary
   spec.homepage      = "https://relishapp.com/vcr/vcr/docs"
-  spec.license       = "MIT-Hippocratic-2.0" 
+  spec.license       = "MIT-Hippocratic-2.0"
 
   spec.files         = Dir[File.join("lib", "**", "*")]
   spec.executables   = Dir[File.join("bin", "**", "*")].map! { |f| f.gsub(/bin\//, "") }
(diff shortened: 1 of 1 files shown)


vcr.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |spec|
   spec.summary       = %q{Record your test suite's HTTP interactions and replay them during future test runs for fast, deterministic, accurate tests.}
   spec.description   = spec.summary
   spec.homepage      = "https://relishapp.com/vcr/vcr/docs"
-  spec.license       = "MIT-Hippocratic-1.2"
+  spec.license       = "MIT-Hippocratic-2.0"
```
olleolleolle:
```suggestion
  spec.license       = "MIT-Hippocratic-2.0"
```

vcr.gemspec
```diff
@@ -13,7 +13,7 @@ Gem::Specification.new do |spec|
...
```
pombredanne: Why do you call this the MIT-hippocratic-2.0?
This is officially called at https://firstdonoharm.dev/ "The Hippocratic License 2.0"
So a tag should be best be "hippocratic-2.0" rather than the possibly confusing "mit-hippocratic-2.0" ...
@CoralineAda would you concur?

CoralineAda:
Yes, it’s not MIT-adjacent anymore.

vcr.gemspec
pombredanne: Note that If you still have MIT-licensed code too, then you should consider using an SPDX-like license expression syntax such as: `MIT AND hippocratic-2.0`

krainboltgreene: Since RubyGems doesn't have a spec for this, I suppose any will do.

PR #815, opened 2020-04-11
Remove multi_json dependency, yajl-ruby use only JSON

### LINKED ISSUE #789: VCR doesn't declare its dependency on multi_json
It seems this was discussed in #408 almost 6 years ago and it has not changed. It would be nice to add a special error message if it couldn't be loaded.

Ruby 2.5.3p105
Gem 5.0.0
Rails 5.2.2.1
Rspec 3.9.0

olleolleolle:
This ~is a WIP PR to explore~ does "stop using `multi_json`" as described in #789.

Fixes #789

## Steps

- remove the dependency
- find errors in breaking tests
- also, remove the now-unused yajl-ruby from the `Gemfile`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,3 @@ platform :mri do
   gem "curb", "~> 0.9.0"
 end
 
-platform :ruby do
-  gem "yajl-ruby"
-end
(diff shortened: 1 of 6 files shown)


olleolleolle:
(Can the dependency `yajl-ruby` also be dropped, with this?)

olleolleolle:
I think this is ready to go in.

Issue #687, opened 2018-01-23
Expired cassettes allowed to record with record: :none


Billiam:
If a cassette has previously been recorded, but has expired, setting the record mode to `:none` still allows requests to go through, and updates the cassette (in versions 3.0.3 and 4.0.0).

```ruby
VCR.use_cassette('example', re_record_interval: 1.second) do
    Net::HTTP.get_response('example.com', '/bar')
end
sleep 5
VCR.use_cassette('example', record: :none, re_record_interval: 1.second) do
    Net::HTTP.get_response('example.com', '/bar')
end
```
**Expectation:** Example cassette is initially recorded, and the second attempt to use the (now expired) cassette when the record mode is `:none` raises an `UnhandledHTTPRequestError`

**Behavior:** Both requests succeed, and the example request is updated.

coding-bunny:
I know this issue is old, but I think I'm running into this problem as well now, when using `Timecop` to freeze a test and having the `re_record_interval` set. Even when running with the `:none` option, VCR keeps updating the cassettes.

And the response part is being updated, as the requests themselves don't change.

coding-bunny:
So I changed our VCR settings by removing `:re_record_interval` and setting the record mode to `:once` and now VCR is no longer updating the cassettes but actually re-using them.

I feel when using Timecop to freeze time for specific requests, VCR really doesn't play nice.

PR #824, opened 2020-06-19
Fix: the behavior of the option re_record_interval "none"

nicolasiensen:
In this pull request, we are changing how the option `re_record_interval` behaves.

If this option is provided together with `record` mode set to `none`, VCR won't re-record cassettes, even if their intervals have expired.

This commit solves the issue #687


nicolasiensen:
Hey @olleolleolle, what is the process to get the PR merged? Am I missing something?

PR #832, opened 2020-09-24
Fix compatibility with frozen string literals

casperisfine:
We're running one of our app's CI with `RUBYOPT="--enable-frozen-string-literal"`, and VCR fails with:

```
TypeError: can't define singleton
gems/vcr-3.0.3/lib/vcr/version.rb:15:in `version'
gems/vcr-3.0.3/lib/vcr/errors.rb:78:in `relish_version_slug'
```


olleolleolle (approved):
Thanks for the fix!

PR #833, opened 2020-09-25
Transforms ERB hash keys to symbol, in case string

z1lk:
Passing an ERB hash with string keys was causing a NameError in
struct cache initialization, as Struct members must be symbols.
The NameError was being rescued and re-raised, saying the cassette
"references undefined variable", which was misleading.

### DIFF
--- a/lib/vcr/cassette/erb_renderer.rb
+++ b/lib/vcr/cassette/erb_renderer.rb
@@ -40,8 +40,9 @@ module VCR
       end
 
       @@struct_cache = Hash.new do |hash, attributes|
-        attributes.map!(&:to_sym)
-        hash[attributes] = Struct.new(*attributes)
+        attributes = attributes.map(&:to_sym)
+        hash[attributes] = Struct.new(*attributes) unless hash.key?(attributes)
+        hash[attributes]
       end
 
       def variables_object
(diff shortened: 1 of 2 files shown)


lib/vcr/cassette/erb_renderer.rb
```diff
@@ -40,6 +40,7 @@ def template
       end
 
       @@struct_cache = Hash.new do |hash, attributes|
+        attributes.map!(&:to_sym)
         hash[attributes] = Struct.new(*attributes)
       end
```
krainboltgreene, line 45:
```suggestion
        hash[attributes] = Struct.new(*attributes.map(&:to_sym))
      end
```

krainboltgreene (changes requested):

lib/vcr/cassette/erb_renderer.rb
```diff
@@ -40,8 +40,9 @@ def template
...
 
       @@struct_cache = Hash.new do |hash, attributes|
-        attributes.map!(&:to_sym)
-        hash[attributes] = Struct.new(*attributes)
+        attributes = attributes.map(&:to_sym)
+        hash[attributes] = Struct.new(*attributes) unless hash.key?(attributes)
+        hash[attributes]
```
z1lk, line 45: @krainboltgreene Your suggested change got me thinking about reducing the number of keys in the cache. All of the following end up creating Structs with the same members and should reuse the cached Struct rather than create a new one:

```ruby
@@struct_cache[ [:a, :b] ]
@@struct_cache[ [:a, 'b'] ]
@@struct_cache[ ['a', :b] ]
@@struct_cache[ ['a', 'b'] ]
```

krainboltgreene (approved):

PR #836, opened 2020-10-03
Group cassettes

### LINKED ISSUE #700: Add :single_cassette option to RSpec metadata

hirowatari:
Fixes #700, this will hopefully close that and add the :single_cassette option.  See also #761 for a previous pull request.

Some questions I have:
- I think documentation is necessary for this, but was unsure where it should go.  I had considered putting it in `features/test_frameworks/rspec_metadata.feature`.
- I believe that this feature would not apply outside of rspec.  Is that correct?  I have only used VCR in the context of rspec
- I think `unique_cassette` might be a better name, but there didn't seem to be any consensus in #700, so I just left it.


hirowatari:
Darn I don't understand why this is failing.  Can a maintainer chime in?  It works on my local with the same ruby version and same seed so I'm not sure what's up.

olleolleolle:
```
  4) VCR::Cassette::HTTPInteractionList has_interaction_matching? does not consume the interactions when they match
     Failure/Error: if VCR.configuration.preserve_exact_body_bytes_for?(self)
       #<Double "VCR::Configuration"> received unexpected message :preserve_exact_body_bytes_for? with (#<struct VCR::Request method=:post, uri=nil, body="", headers={}>)
```

This was the common test failure. InteractionList tests.

mcmire:
Not to bikeshed, but just throwing some more ideas for naming out there:

* I feel like `use_single_cassette` (instead of `single_cassette`) would be more consistent with the existing configuration values. That said, I don't know if the implication that the _requests_ would be put under a single cassette is obvious enough.
* I'm actually not completely happy with `group_cassettes` either — does that mean cassettes are grouped in one folder or that the requests that would have shown up in individual cassettes are grouped in one? A better name would be `group_requests`, although, like `use_single_cassette`, it's not completely obvious _where_ or _how_ they'd be grouped.
* What about `consolidate_cassettes`? It's very succinct but I feel like consolidation implies that there is a singular location where everything will be placed, and if you combine multiple cassettes together, cassettes are composed of requests, so by definition you're also combining the requests together too.

PR #837, opened 2020-10-03
Update docs to warn of a possible gotcha

### LINKED ISSUE #717: Filtering of phrases with special characters fails
I got a configuration:

```ruby
VCR.configure do |c|
  c.filter_sensitive_data("<FILTERED>") { SomeService.password }
end
```

It happens that password contains "&" character, like "some&password".
It's not filtered, as written value in the cassette is `some&amp;password` and it doesn't match with the source.

Workaround for the problem was `CGI::escapeHTML(SomeService.password)`

hirowatari:
`filter_sensitive_data` may silently ignore replacing
the intended text if it has been encoded.

fixes #717

### DIFF
--- a/features/configuration/filter_sensitive_data.feature
+++ b/features/configuration/filter_sensitive_data.feature
@@ -26,8 +26,11 @@ Feature: Filter sensitive data
 
   Warning, the data you're filtering may have been escaped or encoded.
   Thus you may need to escape the sensitive text. For example, if
-  "some&password" is in your cassette, you will need to filter
-  "some&amp;password". You can also use `CGI::escapeHTML('some&password')`.
+  "some&password" is in your cassette within a context which is html
+  escaped, you will need to filter "some&amp;password" or
+  `CGI::escapeHTML('some&password')`. Similarly if it's escaped in a
+  query string, then you will need to filter "some%26password"
+  or `CGI::escape('some&password')`
 
   You can specify as many filterings as you want.
 
--- a/features/configuration/filter_sensitive_data.feature
+++ b/features/configuration/filter_sensitive_data.feature
@@ -24,6 +24,11 @@ Feature: Filter sensitive data
   substitution string so that the interaction will be identical to what was
   originally recorded.
 
+  Warning, the data you're filtering, may have be escaped or encoded.
+  Thus you may need to escape the sensitive text. For example if
+  "some&password" is in your cassette you will need to filter
+  "some&amp;password". You may wish to use `CGI::escapeHTML('some&password')`.
+
   You can specify as many filterings as you want.
 
   Scenario: Multiple filterings
--- a/features/configuration/filter_sensitive_data.feature
+++ b/features/configuration/filter_sensitive_data.feature
@@ -24,10 +24,10 @@ Feature: Filter sensitive data
   substitution string so that the interaction will be identical to what was
   originally recorded.
 
-  Warning, the data you're filtering, may have be escaped or encoded.
-  Thus you may need to escape the sensitive text. For example if
-  "some&password" is in your cassette you will need to filter
-  "some&amp;password". You may wish to use `CGI::escapeHTML('some&password')`.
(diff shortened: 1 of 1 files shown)


features/configuration/filter_sensitive_data.feature
```diff
@@ -24,6 +24,11 @@ Feature: Filter sensitive data
   substitution string so that the interaction will be identical to what was
   originally recorded.
 
+  Warning, the data you're filtering, may have be escaped or encoded.
```
nicolasiensen, line 27:
```suggestion
  Warning, the data you're filtering may have been escaped or encoded.
```

features/configuration/filter_sensitive_data.feature
```diff
@@ -24,6 +24,11 @@ Feature: Filter sensitive data
...
+  Thus you may need to escape the sensitive text. For example if
```
nicolasiensen, line 28:
```suggestion
  Thus you may need to escape the sensitive text. For example, if
```

features/configuration/filter_sensitive_data.feature
```diff
@@ -24,6 +24,11 @@ Feature: Filter sensitive data
...
+  "some&password" is in your cassette you will need to filter
```
nicolasiensen, line 29:
```suggestion
  "some&password" is in your cassette, you will need to filter
```

features/configuration/filter_sensitive_data.feature
```diff
@@ -24,6 +24,11 @@ Feature: Filter sensitive data
...
+  "some&amp;password". You may wish to use `CGI::escapeHTML('some&password')`.
```
nicolasiensen, line 30:
```suggestion
  "some&amp;password". You can also use `CGI::escapeHTML('some&password')`.
```

nicolasiensen (commented):
I just have some wording suggestions. 🤓

hirowatari:
Thanks for the suggestions.  I've made the requested changes.

nicolasiensen (approved):

features/configuration/filter_sensitive_data.feature
```diff
@@ -24,6 +24,11 @@ Feature: Filter sensitive data
   substitution string so that the interaction will be identical to what was
   originally recorded.
 
+  Warning, the data you're filtering may have been escaped or encoded.
+  Thus you may need to escape the sensitive text. For example, if
+  "some&password" is in your cassette, you will need to filter
+  "some&amp;password". You can also use `CGI::escapeHTML('some&password')`.
```
timdiggins, line 30: Great idea and really helpful (I was stuck with this for a while).

It might be good to add that query params need to use `CGI::escape` rather than `CGI::escapeHTML`

olleolleolle: @timdiggins Cool, do you have a wording example, or would you like to make a follow-up PR?

timdiggins: Good point @olleolleolle. Here's a propsal. Looks like this branch is now out of date, so may make sense for @hirowatari to do all the changes, if agreeable. I can do a follow-up PR if needed.

    Warning, the data you're filtering may have been escaped or encoded.
    Thus you may need to escape the sensitive text. For example, if 
    "some&password" is in your cassette within a context which is html 
    escaped, you will need to filter "some&amp;password" or 
    `CGI::escapeHTML('some&password')`. Similarly if it's escaped in a 
     query string, then you will need to filter "some%26password" 
     or `CGI::escape('some&password')`

timdiggins: PS could leave out the last sentence if the text is now too long, it's more the "which is html escaped" that helps indicate that escapeHTML is only one of the possible escaping solutions

olleolleolle (approved):
Thanks @hirowatari! Your attention to detail and helpful documentation fix helps others a lot!

PR #838, opened 2020-10-03
Allow a Host trailing dot to match same Host without it

### LINKED ISSUE #681: VCR doesn't match URLs ending with a single dot (.)
### Background
We use Docker containers for local development, and the hostnames of those containers end with `.`, for example `http://www.:10000` or `http://api.:10010`. This doesn't seem to play well with VCR.

### To Reproduce
To reproduce this, you can run any server locally and add this line to your /etc/hosts:
```
127.0.0.1 example.
```
With a Rails server running on port 10000, I run the following script:
```
require 'vcr'
require 'net/http'

VCR.configure do |config|
  config.cassette_library_dir = "fixtures/vcr_cassettes"
  config.hook_into :webmock
end

VCR.use_cassette("missing_dot") do
  Net::HTTP.get_response(URI('http://example.:10000/'))
end
```
VCR generates a cassette with this at the top:
```
---
http_interactions:
- request:
    method: get
    uri: http://example:10000/
...
```
Running the script again works, but the recorded URI dropped the `.`, recording `example:10000` instead of `example.:10000`.

If I manually change the URI in the cassette to properly include the `.` and run the script again, I receive the following error:
```
================================================================================
An HTTP request has been made that VCR does not know how to handle:
  GET http://example:10000/
...
```

### Notes
I suspected a WebMock error, but this script matches fine:
```
require 'webmock'
include WebMock::API

WebMock.enable!

stub_request(:get, "http://example.:10000/")
Net::HTTP.get("example.:10000", "/")
```
Although if I remove the `stub_request(...)` line and just do the get request without mocking first, WebMock suggests 
```
You can stub this request with the following snippet:

stub_request(:get, "http://example:10000/")
...
```
which will also match fine, despite missing the `.`.

hirowatari:
Fixes #681. See for further comments.

### DIFF
--- a/lib/vcr/request_matcher_registry.rb
+++ b/lib/vcr/request_matcher_registry.rb
@@ -110,12 +110,18 @@ module VCR
 
     def register_built_ins
       register(:method)  { |r1, r2| r1.method == r2.method }
-      register(:uri)     { |r1, r2| r1.uri == r2.uri }
+      register(:uri) do |r1, r2|
+        uri1 = URI(r1.uri)
+        uri2 = URI(r2.uri)
+        uri1.host = uri1.host.chomp('.')
+        uri2.host = uri2.host.chomp('.')
+        uri1 == uri2
+      end
       register(:body)    { |r1, r2| r1.body == r2.body }
       register(:headers) { |r1, r2| r1.headers == r2.headers }
 
       register(:host) do |r1, r2|
-        r1.parsed_uri.host == r2.parsed_uri.host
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Thanks for making these equivalent!

Makes sense, and the implementation is straightforward and small.

olleolleolle:
Ah, @hirowatari, this is super. 

Can you add a line to the Changelog about this?

hirowatari:
Thanks @olleolleolle.  Updated the changelog.

PR #845, opened 2021-01-24
Support Cucumber-Ruby v4 and later

### LINKED ISSUE #843: Cucumber 5.2.0 no longer supports the .feature method to get the name of a test case, used in "use_scenario_name"
The VCR method https://github.com/vcr/vcr/blob/v6.0.0/lib/vcr/test_frameworks/cucumber.rb#L51-L52 still uses the "feature" method, which is no longer supported by cucumber 5.2.0.
Are there any plans to fix this issue?

brasmusson:
In version 4 of Cucumber-Ruby the execution of test cases is based on Pickles. Therefore less data from the feature files are available to the before/after/etc hooks. Also the Gherkin grammar has changed so that (all) Scenarios can have Examples - that is "Scenario Outline" is not just synonym for "Scenario".

Therefore, in case of the "use scenario name" option, use the feature file name instead of feature name, and use the row of the example to create uniqueness for Scenarios with Examples.

Fixes #843.

### DIFF
--- a/lib/vcr/test_frameworks/cucumber.rb
+++ b/lib/vcr/test_frameworks/cucumber.rb
@@ -47,9 +47,20 @@ module VCR
                 scenario.scenario_outline.name,
                 scenario.name.split("\n").first
               ].join("/")
-            else
+            elsif scenario.respond_to?(:feature)
               [ scenario.feature.name.split("\n").first,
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
Thank you, @brasmusson!

PR #881, opened 2021-03-02
Switch to GitHub Actions

bradshjg:
Previous discussion at #851

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ Changelog
 =========
 
 ## Unreleased changes
+[Full Changelog](https://github.com/vcr/vcr/compare/v6.0.0...master)
 
 - [breaking] JSON-serializer generates pretty-formatted output.
 - [new] Add `VCR.turned_on` similar to `VCR.turned_off` (#681)


script/cached-bundle
```diff
@@ -1,48 +0,0 @@
...
-
-bundle "$@"
-BUNDLE_GEMFILE=gemfiles/faraday_old.gemfile bundle "$@" --path "$PWD"/vendor/bundle
-BUNDLE_GEMFILE=gemfiles/typhoeus_old.gemfile bundle "$@" --path "$PWD"/vendor/bundle
-
-if [ ! -f "$cache_name" ] && [ -n "$AMAZON_SECRET_ACCESS_KEY" ]; then
-  echo "Caching \`${bundle_path}' to S3"
-  tar czf "$cache_name" "$bundle_path"
-  script/s3-put "$cache_name" "${AMAZON_S3_BUCKET}:${TRAVIS_REPO_SLUG}/${cache_name}"
-fi
```
krainboltgreene, line 48: Christ, is there just an S3 bucket somewhere with a whole bunch of code?

bradshjg: 😄 There might still be, but I think the script is vestigial from before TravisCI offered native caching (https://mislav.net/2013/12/travis-cached-bundle/ appears to be the inspiration).

krainboltgreene:
So I want to merge this as a part 1 for a 2 part series. A lot of the script stuff is really just a hack to get what we have in github actions today, for example spinning up a http server.

bradshjg:
Am I understanding correctly that for this PR we want the existing behavior where an app is spun up on an another thread in https://github.com/vcr/vcr/blob/master/spec/support/vcr_localhost_server.rb#L30 but a future PR might push that out to a [service container](https://docs.github.com/en/actions/guides/about-service-containers)?

Let me know if there's any way I can help out with fixing up this PR or helping out in the future. Getting to know this codebase is helpful as I'm looking into building a openapi spec based generative load testing library (that's a mouthful) where I'll need to do a lot of similar things (well, except for the hard parts y'all have 😄 ).

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+
+    strategy:
+      matrix:
+        ruby-version: [2.4, 2.5, 2.6, 2.7]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install OS dependencies
+      run: |
+        sudo apt-get install libcurl4-openssl-dev
```
olleolleolle, line 22:
```suggestion
        sudo apt-get install --assume-yes libcurl4-openssl-dev
```

Perhaps we want something like that for it be non-interactive?

> One side effect of using `-y` is that it will skip downgrades, whereas if the user entered y, it would perform the downgrades. So if you want `-y` to execute the same thing as a user entering y, you also need to add `--allow-downgrades`.

But we don't need downgrades.

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
+name: Ruby CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
olleolleolle, line 7:
```suggestion
  pull_request:
    types: [opened, synchronize, reopened]
```

If we added that, we'd run the job for all commits, for all PRs.

I searched around, and found this longer explanation of those values:
https://github.community/t/trigger-workflow-for-a-pull-request-and-for-all-subsequent-commits-in-it/17996

.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  test:
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        ruby-version: [2.4, 2.5, 2.6, 2.7]
```
olleolleolle, line 16: Can we add in `"3.0"` as well here?

olleolleolle (approved):
I am in favor of trying this out step by step, since the Travis Pull Request webhook is gone, already.

EDIT: Let's make that "let us experiment until we get to 🟢 green builds in this PR".

Mingled with the output, I can see that with Faraday, there are messages about its backends not being loaded

```
  cannot load such file -- patron
```

and about em-http-request, another such gem being a Faraday backend.

Perhaps the Travis setup CI scripts were taking care to get those installed?

.github/workflows/test.yml
bradshjg: I ended up merging this down to the default branch in my fork because I couldn't figure out if I could run tests from a branch if I was just creating a `.workflows` for the first time. I think I was wrong about that being necessary, but more importantly...

Is this PR being based on my default branch preventing y'all from making any changes? There's a checkbox that does say it should allow edits...but I'm not sure if that actually works if I'm using the default branch in my fork. Apologies if that's the case and I'm happy to resubmit from a branch if that would help y'all.

olleolleolle: We just have limited bandwidth.

script/cached-bundle
krainboltgreene: Got confirmation from the original author that the bucket is gone.

.github/workflows/test.yml
bradshjg: Totally understand, no rush on my part, just wanted to make sure I wouldn't be a bottleneck.

I should say in the spirit of transparency that in between opening this PR and now, I started working at GitHub. I don't work on anything related to Actions and don't think there's any conflict between working there and helping this PR land, but wanted to let y'all know.

bradshjg: With respect to types, my reading of https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request is that the types you've listed are the default, but it's not bad to be explicit.

I also think it's not a bad idea to remove the limit of only running for PRs where the master branch is part of it, all PRs deserver CI 😄

.github/workflows/test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby-version: [2.4, 2.5, 2.6, 2.7]
+        ruby-version: [2.4, 2.5, 2.6, 2.7, 3.0]
```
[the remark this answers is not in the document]
bradshjg: Nice catch, thanks! Should be fixed now.

bradshjg:
I'll take a look this weekend to see if I can get this green again.

vcr.gemspec
```diff
@@ -42,4 +42,9 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "relish"
   spec.add_development_dependency "mime-types"
   spec.add_development_dependency "sinatra"
+  ruby_version = RUBY_VERSION.split('.').map { |x| x.to_i } # TODO what about prereleases?
+  if (ruby_version <=> [3, 0, 0]) >= 0
```
olleolleolle, line 46: Perhaps we can check "does it start with 3"?

bradshjg: I like that idea thanks, you could probably tell that I thought this would be a "quick fix" and tested things in CI instead of setting up a local testing environment. I'm going to fix that because hopefully it was just me...but 50 failing CI e-mails were quite unexpected 😆.

I'm also going to see if I can fix all the traceback noise in the builds because they're very hard to parse right now even when builds pass (which they do for most Ruby versions).

PR #882, opened 2021-03-18
extract vcr_cassette_name_for

dabroz:
I would like to use 

```ruby
    #   VCR.use_cassette("some cassette") do |cassette|
    #     Timecop.freeze(cassette.originally_recorded_at || Time.now) do
    #       # ...
    #     end
    #   end
```

but since I am doing the `Timecop.freeze` as a RSpec `around` hook, the cassette is not loaded by then. I can get the originally_recorded_at from the cassette manually but I need to know the cassette name/path. `vcr_cassette_name_for` is currently not really accessible externally and this PR would make life much easier. :)

### DIFF
--- a/lib/vcr/test_frameworks/rspec.rb
+++ b/lib/vcr/test_frameworks/rspec.rb
@@ -5,27 +5,28 @@ module VCR
     module Metadata
       extend self
 
+      def vcr_cassette_name_for(metadata)
+        description = if metadata[:description].empty?
+                        # we have an "it { is_expected.to be something }" block
+                        metadata[:scoped_id]
+                      else
+                        metadata[:description]
+                      end
+        example_group = if metadata.key?(:example_group)
+                          metadata[:example_group]
+                        else
+                          metadata[:parent_example_group]
+                        end
+
+        if example_group
+          [vcr_cassette_name_for(example_group), description].join('/')
+        else
+          description
+        end
+      end
+
       def configure!
         ::RSpec.configure do |config|
-          vcr_cassette_name_for = lambda do |metadata|
-            description = if metadata[:description].empty?
-                            # we have an "it { is_expected.to be something }" block
-                            metadata[:scoped_id]
-                          else
-                            metadata[:description]
-                          end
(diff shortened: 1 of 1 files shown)


lib/vcr/test_frameworks/rspec.rb
```diff
@@ -5,27 +5,28 @@ module RSpec
     module Metadata
       extend self
 
+      def vcr_cassette_name_for(metadata)
+        description = if metadata[:description].empty?
```
olleolleolle, line 9: Minor: Perhaps we can use a newline after the assignment `=`, to reduce the amount of indentation used?

```ruby
description =
  if meta....
    # ...
  else 
    # ...
  end
```

dabroz: Sure thing.

olleolleolle: Thanks!

lib/vcr/test_frameworks/rspec.rb
```diff
@@ -5,27 +5,30 @@ module RSpec
...
+            metadata[:scoped_id]
+          else
+            metadata[:description]
+          end
+        example_group = 
+          if metadata.key?(:example_group)
+            metadata[:example_group]
+          else
+            metadata[:parent_example_group]
+          end
```
olleolleolle, line 21: Would this ugly thing do the same?

```ruby
example_group = metadata.fetch(:example_group, metadata[:parent_example_group])
```

olleolleolle (approved):
I got curious about why it was sealed into a lambda inside a block, couldn't find any specific reason.

A regular method looks nice.

dabroz:
Yeah, lambda wouldn't be my first choice here, but perhaps there was a reason for this before.

Would you mind releasing a new version with this? Would help a lot. :)

lib/vcr/test_frameworks/rspec.rb
krainboltgreene: Change the coding style.

krainboltgreene (approved):

PR #814, opened 2020-03-31
Allow Excon >= 0.62.0

olleolleolle:
This PR starts allowing Excon **`>= 0.62.0`** instead of `= 0.62.0`.

This is a follow-up from #809 

---

## Background Notes

This PR is a work in progress. The ongoing notes are left here as aide-mémoires.

The current test which fails (intermittently) is 

> rspec ./spec/acceptance/concurrency_spec.rb:15 # VCR when used in a multithreaded environment with an around_http_request can use a cassette in an #around_http_request hook

### the Excon warning `Invalid Excon connection keys: :remote_ip`

`v0.73.0` worked for most of the tests: is this a flake?

```
[excon][WARNING] Invalid Excon connection keys: :remote_ip
```

[excon: search for "remote_ip"](https://github.com/excon/excon/search?q=remote_ip&unscoped_q=remote_ip) -- perhaps something about its mocks?

The `Response` class has a setter and getter for it. Into its `@data` member.
https://github.com/excon/excon/blob/d59f34094e195cae39cd30c019c4130080cb8f4c/lib/excon/response.rb#L41-L46

This `mock.rb`:
https://github.com/excon/excon/blob/v0.73.0/lib/excon/middlewares/mock.rb#L29

**Update**: Attempted a CI retry, by `git commit --amend --no-edit`

PR #883, opened 2021-04-11
GH Actions CI

bradshjg:
Supercedes https://github.com/vcr/vcr/pull/881
Related to https://github.com/vcr/vcr/discussions/851

This represent a fairly drastic change in how tests are run, previously a single script `ci.sh` was responsible for running all the steps and appropriately timing and folding the relevant sections. This PR splits the CI steps into separate Actions steps, ~~but doesn't preserve at least one (probably important!) behavior: testing for warnings.~~ Edit: this now does check for warnings but the semantics might differ slightly from `ci.sh`.

~~It's not clear to me whether we want to dump all warnings in various steps to a file or create a similar abstraction as `ci.sh` had where warning checks were sorta built-in to the various steps. I found the indirection in `ci.sh` painful to follow and personally would probably prefer more explicit (if duplicated) `2> {some_warnings_file}` and then a step to crawl through those at the end, but I'm totally open to suggestions.~~ Edit: each test step dumps stderr to a file and then that file is checked for warnings in a later step.

Other potentially controversial things:

* `em-http-request` adds headers by default, and in the test suite there's very reasonable header tests that get broken by this behavior. I opted to update the request options in a way that omits those headers so the tests pass as written, but it's kinda weird and fragile but hey they started it! 😄 
*  I skipped a flaky multi-threaded test. See https://github.com/vcr/vcr/pull/814 for a little more context.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -64,7 +64,7 @@ jobs:
 
     - name: check warnings
       run: |
-        vcr_warnings="$(grep -F "$PWD" "${ALL_WARNINGS}" | grep "warning: " | sort | uniq -c | tee /dev/stderr | wc -l)"
+        vcr_warnings="$(grep -F "$PWD" "${ALL_WARNINGS}" | grep "warning: " | grep -v "${PWD}/vendor/bundle" | sort | uniq -c | tee /dev/stderr | wc -l)"
         if [ "$vcr_warnings" -gt 0 ]; then echo "FAILED: test suite doesn't tolerate warnings"; exit 1; fi
 
     - name: doc coverage
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,62 @@
+name: vcr ci
+
+on:
+  push:
+    branches: [ $default-branch ]
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    env:
+      RUBYOPT: "-w"
+
+    strategy:
+      matrix:
+        ruby-version: ["2.5", "2.6", "2.7", "3.0"]
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Install OS dependencies
+      run: sudo apt-get install --assume-yes libcurl4-openssl-dev
+
+    - name: Set up Ruby ${{ matrix.ruby-version }}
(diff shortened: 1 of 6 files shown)


bradshjg:
Sorry to open a new PR about this, but the other one was long enough ago that I completely re-imagined how I wanted to do this. I hope this makes sense to y'all too.

Things we talked about that I want to highlight:

> "let us experiment until we get to 🟢 green builds in this PR".

That's the spirit I went with, which means stuff like

> Can we add in "3.0" as well here?

I omitted because it is going to break and that probably deserves its own PR.

krainboltgreene:
Okay, so what do we need to get this merged? If we have to strip stuff out that's fine, we can always work on readding those things, but I would love a list of things we're leaving behind.

bradshjg:
> Okay, so what do we need to get this merged? If we have to strip stuff out that's fine, we can always work on readding those things, but I would love a list of things we're leaving behind.

The only thing I know that we're dropping with respect to testing is a test for the vcr library emitting warnings (with careful attention to exiting cleanly if only dependencies emit warnings), but I'd love a second set of 👀 on that.

I also would like to drop the entire `script/` folder before merge, but that's because I like my PRs 🟥 🟥  😄

bradshjg:
@olleolleolle @krainboltgreene gentle bump to see if there's anything I can help to move this forward (and apologies for the direct mentions if they prove less than helpful) 😄

.github/workflows/test.yml
```diff
@@ -0,0 +1,73 @@
+name: vcr ci
+
+on:
+  push:
+    branches: [ $default-branch ]
+  pull_request:
```
olleolleolle, line 6: I'm curious if this alias for default branch does the right thing. It's be cool.

olleolleolle (approved):
In order to keep the momentum, @bradshjg, let's merge this, which added a single defined skip, which can be investigated. Warnings are collected and analyzed. 

Em-http-related testing fixes for reproducibility could potentially be a separate thing, but I'm OK with those spec changes landing at the same time.

I had a tiny note about the $default-branch name, which is easy to see if it worked or not, after a merge. Otherwise, we can change it to the relevant string before merge. Your call, @bradshjg.

In all, this looks like future to me.

olleolleolle (approved):
Clear ahead!

bild

olleolleolle:
@bradshjg Hooray! Thank you so much for your hard work in creating this!

PR #888, opened 2021-05-26
Add :clean record mode for cleaning unused HTTP interactions from cassettes

bazay:
Adds a new record mode :clean to actively check for unused HTTP interactions and automatically removes them from existing cassettes.
This helps reduce cassette file size as well as provide a clear picture of actual requests being made for a cassette.

### Caveats
- Clears unused HTTP interactions in a cassette file. This may have unintended consequences for cassette files used by a number of different examples but only use a subset of the HTTP interactions.

### Additional Notes
- This feature works similarly to the configuration option `clean_outdated_http_interactions` however that requires `re_record_interval` which will have the result of clearing used http interactions that might be old but still valid.

### DIFF
--- a/lib/vcr/cassette.rb
+++ b/lib/vcr/cassette.rb
@@ -14,7 +14,8 @@ module VCR
     #   * :new_episodes -- Playback previously recorded HTTP interactions and record new ones.
     #   * :once -- Record the HTTP interactions if the cassette has not already been recorded;
     #              otherwise, playback the HTTP interactions.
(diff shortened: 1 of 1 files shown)


ryanseys:
https://github.com/vcr/vcr/pull/946 is now merged which implements this functionality, albeit a different approach. 

Not yet released, but we can likely close this.

bazay:
@ryanseys thanks for notifying. Great to have this functionality either way 👍

PR #895, opened 2021-07-01
Add localhost_ignored? to request_ignorer

ThHareau:
In our setup, we want to use VCR in a helper, which needs to ignore localhost. At the end of the helper, we want to unignore localhost, but only if VCR was not configured to ignore localhost before. 

```rb
module SomeHelper
    def action
        with_vcr do 
            call
        end
    end

    private

    def with_vcr(&block)
        localhost_already_ignored = VCR.request_ignorer.localhost_ignored?
        VCR.request_ignorer.ignore_localhost = true unless localhost_already_ignored

        VCR.use_cassette("helper_cassette") { yield block }
        VCR.request_ignorer.ignore_localhost = false unless localhost_already_ignored
    end
end
```

This PR adds the method `localhost_ignored?`.

### DIFF
--- a/spec/lib/vcr/request_ignorer_spec.rb
+++ b/spec/lib/vcr/request_ignorer_spec.rb
@@ -62,7 +62,7 @@ module VCR
     end
 
     context 'when ignore_localhost is set to false' do
-      before(:each) { subject.ignore_localhost = false }
+      before { subject.ignore_localhost = false }
 
       it 'localhost_ignored is false' do
         expect(subject.localhost_ignored?).to eq(false)
--- a/spec/lib/vcr/request_ignorer_spec.rb
+++ b/spec/lib/vcr/request_ignorer_spec.rb
@@ -55,6 +55,18 @@ module VCR
       RequestIgnorer::LOCALHOST_ALIASES.each do |host|
         it_behaves_like "#ignore?", "http://#{host}/foo", true
       end
+
+      it 'localhost_ignored is true' do
+        expect(subject.localhost_ignored?).to eq(true)
+      end
+    end
+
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
This is a helpful addition to this object. It wraps access to the set of ignored hosts and the constant, in a useful name. Thanks for adding this.

ThHareau:
You are welcome :)

spec/lib/vcr/request_ignorer_spec.rb
```diff
@@ -55,6 +55,18 @@ def request(uri)
...
         it_behaves_like "#ignore?", "http://#{host}/foo", true
       end
+
+      it 'localhost_ignored is true' do
+        expect(subject.localhost_ignored?).to eq(true)
+      end
+    end
+
+    context 'when ignore_localhost is set to false' do
+      before(:each) { subject.ignore_localhost = false }
```
olleolleolle, line 65:
```suggestion
      before { subject.ignore_localhost = false }
```

olleolleolle: By using the default, we make it easier to read.

ThHareau: I didn't know that. Should I correct the other in this file as well? (ex: line 52)

olleolleolle: Just do this one, thanks. Other PRs can polish other things.

ThHareau: done!

olleolleolle (approved):
Looks good to me!

PR #900, opened 2021-08-30
CI: Add ruby 3.0 build matrix, and remove 2.4

aka47:
Ruby 3.0 got released some time ago, lets make sure it works in all cases with vcr


.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        ruby-version: ["2.4", "2.5", "2.6", "2.7"]
+        ruby-version: ["2.4", "2.5", "2.6", "2.7", "3.0"]
```
olleolleolle, line 19: Thanks for adding the 3.0!

Perhaps we can order the list in 3.0-2.x order?

aka47: yes, sure.. pushed it.

comment_blueprint.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
```
olleolleolle, line 1: Perhaps this file was not... meant to be introduced?

aka47:
@olleolleolle any ideas on those 2 failing cucumber specs. I am missing a hint right now how this connects.

They are.. 
- bundle exec cucumber features/http_libraries/net_http.feature:143 && features/test_frameworks/test_unit.feature:6

olleolleolle:
> @olleolleolle any ideas on those 2 failing cucumber specs. I am missing a hint right now how this connects.
> 
> They are..
> 
>     * bundle exec cucumber features/http_libraries/net_http.feature:143
> 
>     * features/test_frameworks/test_unit.feature:6

Do they fail in isolation, locally?

aka47:
> > @olleolleolle any ideas on those 2 failing cucumber specs. I am missing a hint right now how this connects.
> > They are..
> > ```
> > * bundle exec cucumber features/http_libraries/net_http.feature:143
> > 
> > * features/test_frameworks/test_unit.feature:6
> > ```
> 
> Do they fail in isolation, locally?

Yes - they also fail locally on my mac. Not for ruby 2.7 - but with ruby 3..

olleolleolle:
@aka47 Then it looks a bit like the Webrick server never starts correctly

aka47:
How about removing ruby 2.4 support for new releases? Nobody should use it anyways.. this is the warning I get when I install it.. 

```
WARNING: ruby-2.4.10 is past its end of life and is now unsupported.
It no longer receives bug fixes or critical security updates.
```

olleolleolle:
Removing 2.4 support would be good.

README.md
```diff
@@ -90,12 +90,12 @@ VCR follows the principles of [semantic versioning](https://semver.org/). The [A
...
-  * MRI 2.5
-  * MRI 2.6
+  * MRI 3.0
   * MRI 2.7
-
-Note that as of VCR 6, only >= 2.4 is explicitly supported.
+  * MRI 2.6
+  * MRI 2.5
+  
+Note that as of VCR 6, only >= 2.5 is explicitly supported.
```
olleolleolle, line 98: This comment is not precise: We have released 6.0.0, which _had_ 2.4 support.

Here is a note 

```suggestion
Note that as of after VCR 6.0.0, only >= 2.5 is explicitly supported. 6.0.0 supported >= 2.4.
```

olleolleolle (commented):
Nice! We have gotten to green. I retitled, and added a small README change for precision.

aka47:
Great. Lets squash and merge it then... :)

I would suggest to upgrade cucumber to 7.0 and remove the outdated typhoeus 0.4.0  patch and testing -  ( it is from June 08, 2012) ) - does that make sense?

olleolleolle:
@aka47 I agree with both efforts as separate Issues we would like to address!

PR #901, opened 2021-08-30
CI: Rely on ruby/setup-ruby@v1's bundle install

olleolleolle:
This PR removes a hopefully-redundant `bundle install`, since `ruby/setup-ruby`'s `bundler-cache: true` does a Bundler installation.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,14 +32,12 @@ jobs:
 
     - name: typhoeus-0.4
       run: |
-        bundle install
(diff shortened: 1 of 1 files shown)


aka47:
@olleolleolle  - I believe you need to run bundle install as it sets a different BUNDLER_GEMFILE for each step

olleolleolle:
@aka47 https://github.com/ruby/setup-ruby#matrix-of-gemfiles ah, only supports "at the Job level" configuring!

PR #905, opened 2021-09-01
Drop support for typhoeus 0.4

aka47:
tyhoeus 0.4 was release almost 10 years ago and is long outdated. The current version is 1.4
Lets remove all patches and specs that target this version.


lib/vcr/library_hooks/typhoeus.rb
```diff
@@ -2,137 +2,131 @@
...
-        # @private
-        class << self
-          def vcr_response_from(response)
-            VCR::Response.new \
-              VCR::ResponseStatus.new(response.code, response.status_message),
-              response.headers,
-              response.body,
-              response.http_version,
-              { "effective_url" => response.effective_url }
+        if ::Typhoeus::Request.method_defined?(:encoded_body)
```
olleolleolle, line 67: Q: Is this a part of the "new"/current API?

spec/monkey_patches.rb
```diff
@@ -77,11 +71,11 @@ def disable_all!
   $original_typhoeus_global_hooks = Typhoeus.on_complete.dup
   $original_typhoeus_before_hooks = Typhoeus.before.dup
 elsif defined?(::Typhoeus::Hydra.global_hooks)
```
olleolleolle, line 73: (Can probably drop the if conditional check here, too)

spec/monkey_patches.rb
```diff
@@ -65,18 +65,10 @@ def disable_all!
   end
 end
 
-if defined?(::Typhoeus.before)
```
olleolleolle, line 74: Ah, _this_ conditional is perhaps needed: we have some `Gemfile` variants where the Typheous gem is not defined, I believe.

aka47:
I believe it looks good and can be merged..

olleolleolle (approved):
This drops patches and special support for an ancient version of Typheous.

olleolleolle:
@aka47 Thanks for maintaining this!

PR #907, opened 2021-10-14
Fix for Ruby 3.1: Use fiber-local var for global_hook_disabled_requests

jhawthorn:
Previously the finalizer being registered failed on Ruby trunk because finalizers are passed an argument and the lambda didn't accept one.

I think it's simpler here to just use a fiber-local (we could use a thread-local, but my gut feeling here was to use fiber-local) variable and have our array cleaned up automatically when the thread exits without need of a finalizer.

### DIFF
--- a/lib/vcr/library_hooks/webmock.rb
+++ b/lib/vcr/library_hooks/webmock.rb
@@ -12,8 +12,6 @@ module VCR
     module WebMock
       extend self
 
-      @global_hook_disabled_requests = {}
-
       def with_global_hook_disabled(request)
         global_hook_disabled_requests << request
 
@@ -25,19 +23,12 @@ module VCR
       end
 
       def global_hook_disabled?(request)
-        requests = @global_hook_disabled_requests[Thread.current.object_id]
+        requests = Thread.current[:_vcr_webmock_disabled_requests]
         requests && requests.include?(request)
       end
 
       def global_hook_disabled_requests
-        requests = @global_hook_disabled_requests[Thread.current.object_id]
-        return requests if requests
-
-        ObjectSpace.define_finalizer(Thread.current, lambda {
-          @global_hook_disabled_requests.delete(Thread.current.object_id)
-        })
-
-        @global_hook_disabled_requests[Thread.current.object_id] = []
+        Thread.current[:_vcr_webmock_disabled_requests] ||= []
       end
 
       # @private


olleolleolle (approved):
Looks good to me.

olleolleolle:
@jhawthorn Would you be able to make a CHANGELOG line which summarizes this change?

nehresma:
With the release of Ruby 3.1.0 a few days ago, would it be possible to cut and release a new VCR release that includes this fix?

rainerborene:
+1

olleolleolle:
Followin SemVer, that's a new major, hold on to your hats.

benhutton:
@olleolleolle is there a way to just ship this as a patch release on top of 6.0.0, without all of the other commits? Or is this specifically what is messing up Faraday? Or is VCR's integration with Faraday going to be problematic regardless?

It's just a shame to have this gate an otherwise clean upgrade to Ruby 3.1!

donv:
> @olleolleolle is there a way to just ship this as a patch release on top of 6.0.0, without all of the other commits? Or is this specifically what is messing up Faraday? Or is VCR's integration with Faraday going to be problematic regardless?

You can put this in your `test_helper.rb` after configuring VCR:

```ruby
# FIXME(uwe): Remove when fixed
# https://github.com/vcr/vcr/pull/907/files
module VCR
  class LibraryHooks
    # @private
    module WebMock
      extend self

      def with_global_hook_disabled(request)
        global_hook_disabled_requests << request

        begin
          yield
        ensure
          global_hook_disabled_requests.delete(request)
        end
      end

      def global_hook_disabled?(request)
        requests = Thread.current[:_vcr_webmock_disabled_requests]
        requests && requests.include?(request)
      end

      def global_hook_disabled_requests
        Thread.current[:_vcr_webmock_disabled_requests] ||= []
      end
    end
  end
end
# EMXIF
```

iGEL:
@benhutton

I just created this patch (to be placed in `config/initializers/patch_vcr.rb:
```ruby
# frozen_string_literal: true

if Rails.env.test?
  unless Gem.loaded_specs["vcr"].version == Gem::Version.create("6.0.0")
    raise "This patch is very likely already in the next VCR release after 6.0.0"
  end

  require "vcr/library_hooks/webmock"

  module VCR
    class LibraryHooks
      module WebMock
        def global_hook_disabled?(request)
          requests = Thread.current[:_vcr_webmock_disabled_requests]
          requests && requests.include?(request)
        end

        def global_hook_disabled_requests
          Thread.current[:_vcr_webmock_disabled_requests] ||= []
        end
      end
    end
  end
end
```

@donv Wow, pretty much the same idea, just minutes apart :wink:

danielricecodes:
> @olleolleolle is there a way to just ship this as a patch release on top of 6.0.0, without all of the other commits? Or is this specifically what is messing up Faraday? Or is VCR's integration with Faraday going to be problematic regardless?
> 
> It's just a shame to have this gate an otherwise clean upgrade to Ruby 3.1!

+1

Since this was merged a while ago, there really needs to be a patch release here.  In the meantime:

```ruby
gem 'vcr', github: 'vcr/vcr' #Edge version for Ruby 3.1 support
```

🤷

Justintime50:
Looks to be released as a part of VCR 6.1

coldnebo:
> @benhutton
> 
> I just created this patch (to be placed in `config/initializers/patch_vcr.rb`:

Thanks! I put this in my project, but I'm getting this error now, any ideas?

```
FrozenError:
  can't modify frozen Array: ["/home/app/.gem/gems/actiontext-7.0.4.3/app/helpers", "/home/app/.gem/gems/actiontext-7.0.4.3/app/models"]
# /home/app/.gem/gems/railties-7.0.4.3/lib/rails/engine.rb:575:in `unshift'

```

PR #911, opened 2021-11-18
YAML Serializer: use unsafe_load if available

casperisfine:
Psych 4.0 made `YAML.load` safe by default, meaning it won't deserialize arbitrary types.

This causes us problem because some of our cassette have `OpenSSL::Buffering::Buffer` instances serialized.


olleolleolle (approved):
I guess this makes it compatible, which is good.

casperisfine:
@olleolleolle done!

olleolleolle (approved):

olleolleolle:
@casperisfine Thank you!

casperisfine:
Welcome, thanks for the very timely merge.

PR #913, opened 2021-12-28
Add Ruby 3.1 to the build matrix

andrehjr:
- Also makes VCR::Cassette::Migrator work with newer psych

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Thanks, @andrehjr!

olleolleolle:
@andrehjr if you feel like making another PR, we probably want to drop support for 2.5 now, too! If this sounds like not your thing, thanks so much for taking care of 3.1.

andrehjr:
@olleolleolle leave it to me :) I'll send another PR

PR #916, opened 2021-12-31
Dont attempt to parse a invalid response(nil) when using `hook_into :excon`

### LINKED ISSUE #651: VCR Does Not Support Recording Broken SSL Certs
I tried to use VCR (with excon, if that matters) to record various types of errors so I can make sure a client library handles them correctly.

In the code before serialization, it doesn't seem like TLS issues are considered.

Here's a code snippet that fails:

```ruby
puts VCR.version
# => 3.0.3
VCR.use_cassette('attempt_1', serialize_with: :json) do
  begin
    Excon.get('https://expired.badssl.com/') # SSL cert has expired.
  rescue => e
    # log something here
  end
end
```

The JSON failure is on [this line](https://github.com/vcr/vcr/blob/master/lib/vcr/structs.rb#L494-L498), response is `nil`, and thererefore VCR fails with the following truncated snippet:
```
NoMethodError: undefined method `to_hash' for nil:NilClass
[....]/lib/vcr/structs.rb:497:in `to_hash'
```

andrehjr:
Hi 🙌 

When using `hook_into :excon` and dealing with an internal error from Excon(i.e: OpenSSL::SSL::SSLError) `VCR::Middleware::Excon::Response#error_call` was breaking because it was trying to parse/record a nil response. The error happened [here on Logger#response_summary](https://github.com/vcr/vcr/blob/master/lib/vcr/util/logger.rb#L23), because `response` is nil. 

I think that the correct behavior would be to let Excon throw its own error up the stack as there's nothing recorded yet [here](https://github.com/excon/excon/blob/dc10c8b4a1f7ef3a166352c911c40a3f1027056a/lib/excon/connection.rb#L106-L110) and there's no need for VCR to attempt to read an empty response for this scenario. Please, correct me if I'm wrong.

When using `hook_into :webmock`  the error is thrown and no cassette is recorded. Same things happens here now.

Fixes #651

### DIFF
--- a/spec/lib/vcr/library_hooks/excon_spec.rb
+++ b/spec/lib/vcr/library_hooks/excon_spec.rb
@@ -45,6 +45,26 @@ RSpec.describe "Excon hook", :with_monkey_patches => :excon do
     end
   end
 
+  context "when Excon raises an internal error before body is processed" do
+    let(:excon) { ::Excon.new('https://expired.badssl.com') }
+    let(:ssl_error) { OpenSSL::SSL::SSLError.new("SSL_connect returned=1 errno=0 state=error: certificate verify failed (certificate has expired)") }
+
+    before do
+      allow(excon).to receive(:socket) { raise ssl_error }
+    end
+
+    def make_request
+      VCR.use_cassette('with_errors', :record => :once) do
+        excon.request(method: :get, :expects => [200])
+      end
+    end
+
+    it 'raises SSL errors and does not try to record a cassette' do
+      expect(VCR).to_not receive(:record_http_interaction)
+      expect { make_request }.to raise_error(Excon::Error::Certificate)
+    end
+  end
+
   include_examples "Excon streaming"
 
   context 'when Excon raises an error due to an unexpected response status' do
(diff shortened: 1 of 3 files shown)


spec/lib/vcr/library_hooks/excon_spec.rb
```diff
@@ -45,6 +45,26 @@ def make_request
...
+      allow(excon).to receive(:socket) { raise ssl_error }
+    end
+
+    def make_request
+      VCR.use_cassette('with_errors', :record => :once) do
+        excon.request(method: :get, :expects => [200])
+      end
+    end
+
+    it 'raises SSL erros and does not try to record a cassette' do
```
olleolleolle, line 62: erros -> errors (typo)

olleolleolle (approved):
I think this looks like a great addition!

Thanks for the good context given in the Description. 

Can you add a CHANGELOG line about the change?

spec/lib/vcr/library_hooks/excon_spec.rb
andrehjr: ops! Fixed!

andrehjr:
Done @olleolleolle ! Thanks for the review 🙌

olleolleolle (approved):
LGTM!

PR #919, opened 2022-01-14
build: Ensure Faraday 1 and 2 are separated, to get to green

olleolleolle:
This is a WIP.

- add a `Gemfile.faraday-2.0` which is used with Ruby 2.6 and up
  - add faraday dependencies as gems, there
- disable all usages of Typhoeus


olleolleolle:
@iMacTia Here's where we are now, with this experiment: we got to green.

There is a PR stop supporting Ruby 2.5, perhaps that could go in, too.

olleolleolle:
Fixed in #1024

PR #924, opened 2022-02-17
Handle the potential renaming of `ASCII-8BIT`

casperisfine:
This is a bit of a weird one. I proposed that `ASCII-8BIT` be renamed to `BINARY` https://bugs.ruby-lang.org/issues/18576

And this code came up as a potentially broken by this change. It's still possible that the name change may happen, that's why I'm submitting this PR.

### DIFF
--- a/Upgrade.md
+++ b/Upgrade.md
@@ -275,7 +275,7 @@ be configured globally with a block, or set on individual cassettes:
 ``` ruby
 VCR.configure do |c|
   c.preserve_exact_body_bytes do |http_message|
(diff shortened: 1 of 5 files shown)


olleolleolle:
😍 
Thanks for offering this change! We will keep it in the back pocket until you, @casperisfine, remind us again, if the proposal makes it into a release of Ruby. (Or any other time you consider appropriate.)

casperisfine:
👋

olleolleolle (approved):
Thank you, @casperisfine!

PR #925, opened 2022-02-18
Lock faraday to version 1

casperisfine:
CI is fully broken because of faraday 2 right now:

```
An error occurred while loading ./spec/acceptance/concurrency_spec.rb.
Failure/Error: require 'typhoeus/adapters/faraday'

NameError:
  method `call' not defined in Faraday::Adapter::Typhoeus

        remove_method :call              if method_defined? :call
        ^^^^^^^^^^^^^
# ./vendor/bundle/ruby/3.1.0/gems/typhoeus-1.1.2/lib/typhoeus/adapters/faraday.rb:28:in `remove_method'
# ./vendor/bundle/ruby/3.1.0/gems/typhoeus-1.1.2/lib/typhoeus/adapters/faraday.rb:28:in `<class:Typhoeus>'
```

Seem like Typhoeus doesn't support faraday 2 yet: https://github.com/typhoeus/typhoeus/issues/686

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,6 +15,7 @@ jobs:
       ALL_WARNINGS: "/tmp/all-warnings"
 
     strategy:
+      fail-fast: false
       matrix:
         ruby-version: ["3.1", "3.0", "2.7", "2.6", "2.5"]
 
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem "faraday", ">= 0.9.2"
+gem "faraday", ">= 0.9.2", "< 2"
 
 gemspec
 
--- a/spec/lib/vcr_spec.rb
+++ b/spec/lib/vcr_spec.rb
@@ -34,7 +34,7 @@ RSpec.describe VCR do
 
     it 'forwards the given options to `Cassette#eject`' do
       cassette = insert_cassette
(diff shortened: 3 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -15,6 +15,7 @@ jobs:
       ALL_WARNINGS: "/tmp/all-warnings"
 
     strategy:
+      fail-fast: false
```
casperisfine, line 18: Adding this to see if the failure is limited to 3.1.

I must admit I don't understand what the failure is about, but I'll try to dig.

olleolleolle: Thanks, I want that setting there!

casperisfine:
Ok, so it seems the failure is 3.0 and 3.1, might be related to keyword args.

casperisfine:
Ok, so here's where it fails:

```ruby
        if RUBY_VERSION >= "3"
          # if both arguments end with Hashes, and if one is a keyword hash and the other is not, they don't match
          if Hash === expected_args.last && Hash === actual_args.last
            if !Hash.ruby2_keywords_hash?(actual_args.last) && Hash.ruby2_keywords_hash?(expected_args.last)
              return false
            end
          end
        end
```

`expected_args` is kw, but `actual_args` isn't. Not sure why, but at least I know what to look for now.

casperisfine:
@olleolleolle we're green.

I also opened https://github.com/rspec/rspec-mocks/pull/1461 because that error message is really confusing.

PR #930, opened 2022-03-29
Remove MIT license tag #806 #810

adrianshort:
The license(s) tag in the `.gemspec` file is used for the license(s) that apply to the package as a whole not for licenses that only apply to some portion of it.

https://guides.rubygems.org/specification-reference/#licenses=


olleolleolle:
The interlinking in the title of the PR did not work fully. In order to make that clearer, I will manually link to what is mentioned:

- https://github.com/vcr/vcr/issues/806
- https://github.com/vcr/vcr/pull/810

PR #935, opened 2022-04-24
Clarify typhoeus callbacks tests

nellirx:
* checking what is yielded in a clearer way
* documented suspicious behaviour with on_body callback when running single Request instance multiple times, i.e. request.run (commented with FIXME)

### DIFF
--- a/spec/lib/vcr/library_hooks/typhoeus_spec.rb
+++ b/spec/lib/vcr/library_hooks/typhoeus_spec.rb
@@ -118,6 +118,7 @@ RSpec.describe "Typhoeus hook", :with_monkey_patches => :typhoeus, :if => (RUBY_
     end
 
     it { expect(request.tap { |r| r.on_headers {} }).not_to be_streaming }
+
     it { expect { |b| use_cassette { request.tap { |r| r.on_headers(&b) }.run }}.to yield_with_args(have_attributes(headers: include('Content-Length' => '18'))) }
     it { expect { |b| use_cassette { request.tap { |r| r.on_headers(&b) }.run }}.to yield_with_args(have_attributes(headers: use_cassette { request.tap { |r| r.on_headers {} }.run }.headers)) }
 
--- a/spec/lib/vcr/library_hooks/typhoeus_spec.rb
+++ b/spec/lib/vcr/library_hooks/typhoeus_spec.rb
@@ -193,24 +193,6 @@ RSpec.describe "Typhoeus hook", :with_monkey_patches => :typhoeus, :if => (RUBY_
     end
   end
 
-  context 'when using on_body callback returning :abort' do
-    def make_request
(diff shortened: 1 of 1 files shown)


olleolleolle:
Looks like a warning that can be fixed - to pass the "no Warnings allowed" part of the suite.

spec/lib/vcr/library_hooks/typhoeus_spec.rb
```diff
@@ -110,46 +110,86 @@ def make_request
...
+    context 'when re-using single request instance' do
+      let(:request) { Typhoeus::Request.new("http://localhost:#{VCR::SinatraApp.port}/localhost_test") }
+
+      it { expect(on_body).to have_attributes(body: 'Localhost response') }
+      it { expect { |b| on_body(&b) }.to yield_with_args('Localhost response', have_attributes(body: '')) }
+
+      context 'after recording cassette (first request)' do
+        before { on_body }
+
+        it { expect(on_body).to have_attributes(body: 'Localhost responseLocalhost response') } # FIXME
```
olleolleolle, line 163: Curious about these: is this change exposing the re-recording issue, or causing it? Should the FIXME be repaired in this PR?

olleolleolle (commented):
Cool that you got it passing! Thanks for all the work!

I can see FIXME notes in the specs, are they "to be fixed in this commit" or are they for the future? If they are for the future, I'd like it very much if they had a run-on comment like "FIXME so that X doesn't Y when Z happens", to clarify what the FIXME refers to.

Thanks again for taking this on!

nellirx:
> If they are for the future, I'd like it very much if they had a run-on comment like "FIXME so that X doesn't Y when Z happens", to clarify what the FIXME refers to

@olleolleolle thank you for the feedback! I have clarified what the FIXME refers to.

spec/lib/vcr/library_hooks/typhoeus_spec.rb
nellirx: My intention was to expose what to me looks like an issue.
I'd prefer to address it separately please, if that's alright of course.

nellirx:
@olleolleolle I decided to remove "FIXME" related tests to avoid confusion

1. In original tests we didn't use single instance of request, so, if we keep them that way, I believe no fixing would be necessary just yet;
2. the tests with FIXME pass, which may suggest that it is intentional behaviour, while FIXME suggest the opposite, it might be confusing;
3. I'll have a close look into issue and create separate PR with those tests if I find a way to fix/improve it.

Hope that's alright. Please let me know if any questions/suggestions.

Thank you!

olleolleolle:
@nellirx Could you rebase this on top of latest?

olleolleolle (approved):
Thanks! 

These are complicated tests, they are, but now they are cut up in their constituting parts.

Thanks for doing that!

PR #951, opened 2022-10-30
Fix recompressing response to happen after replacing sensitive data

cgunther:
Using a combination of `recompress_response` and `filter_sensitive_data`, previously VCR would recompress the response first, then try to replace the sensitive data placeholders with the actual value, however once it's recompressed, the placeholders can't be found anymore to replace. Therefore, once your app/library decoded the recompressed response itself, the placeholders were still there, instead of the expected values.

This happened because both actions were handled by `before_playback` hooks. The recompress hook was registered first upon initialization of the config instance, then the sensitive data hook was registered upon calling `filter_sensitive_data` (always after initiating the config instance), leading VCR to run them in order, recompressing then replacing.

To ensure recompressing always happens last, after all `before_playback` hooks run, I added a new specific hook, `recompress_response`, that gets invoked right after invoking the `before_playback` hook. This assumes you'd never want to run a `before_playback` hook after recompressing.

My usecase is that I'm using the braintree gem, which [requires gzipped responses][1], however the response has "sensitive" data that I want to commit with a placeholder since the value might vary across developer machines, so I use `decode_compressed_response` and `filter_sensitive_data` so that my committed cassettes use a placeholder, then combine it with `recompress_response` so the braintree gem can see the responses as gzipped upon playback, with the placeholders replaced.

[1]: https://github.com/braintree/braintree_ruby/blob/7d4182b7407bd66061faacadec54960e6c876f72/lib/braintree/http.rb#L174-L180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 Changelog
 =========
 
+- [fix] Fix recompressing response to happen after replacing sensitive data (#951) by @cgunther
+
 ## 6.4.0 (Dec 11, 2025)
 [Full Changelog](https://github.com/vcr/vcr/compare/v6.3.1...v6.4.0)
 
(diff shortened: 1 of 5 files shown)


olleolleolle:
@cgunther Thanks for this super PR! Great context, clear implementation. No notes!

I did https://github.com/vcr/vcr/pull/1067 in response to the failed CI job.

olleolleolle (approved):
Thank you, @cgunther!

PR #969, opened 2023-03-23
Add where to put setup code

trisha:


olleolleolle:
@trisha Hi and thanks for the PR!

I saw this section of text, which has a note that is very general: "ensure that this section of code is included before any test cases are run."
https://github.com/vcr/vcr/blob/master/features/getting_started.md#configure-it 

Can you work that into this addition to the README, so that we are not overly prescriptive (not all code bases _have_ a spec_helper.rb, but most have something _like_ it).

olleolleolle (approved):
Thanks, let's add this contextualizing information.

PR #994, opened 2023-10-18
Add dependabot for GH Actions

olleolleolle:
This PR was done as a way of poking the CI for another run, I wanted to know whether code changes or some weirdness was up.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


olleolleolle:
So, the tests are failing on `bundle install` in a Step.

The /tmp/all-warnings.txt file's to published, it'd be cool if it were.

mattbrictson:
@olleolleolle is this perhaps an Actions bug due to the fact that the workflow has 2 jobs that are both named `cucumber`?

I don't really understand why there are 2 jobs with the same name and why one specifies `BUNDLE_GEMFILE` and the other doesn't.

```yaml
- name: cucumber
  run: |
    bundle install
    bundle exec cucumber features/test_frameworks/cucumber.feature 2>> "${ALL_WARNINGS}"
  env:
    BUNDLE_GEMFILE: Gemfile.cucumber
```

```yaml
- name: cucumber
  run: |
    bundle install
    bundle exec cucumber features/ 2>> "${ALL_WARNINGS}"
```

olleolleolle:
Duplicate. Closing!

Issue #995, opened 2023-10-18
CI: bundle install in step cucumber fails


olleolleolle:
See https://github.com/vcr/vcr/pull/994 for more details

Not much to go on. The Ruby 2.7 complains of "circular deps". 

<details>

```
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
Using rake 13.0.6
Using base64 0.1.1
Using ruby2_keywords 0.0.5
Fetching bigdecimal 3.1.4
Fetching concurrent-ruby 1.2.2
Fetching connection_pool 2.4.1
Fetching minitest 5.20.0
Installing bigdecimal 3.1.4 with native extensions
Installing connection_pool 2.4.1
Using mutex_m 0.1.2
Using public_suffix 5.0.3
Using childprocess 3.0.0
Using contracts 0.17
Using builder 3.2.4
Fetching middleware 0.1.0
Installing minitest 5.20.0
Installing concurrent-ruby 1.2.2
Using thor 1.2.2
Fetching thread_safe 0.3.6
Installing middleware 0.1.0
Fetching cucumber-tag-expressions 2.0.4
Installing thread_safe 0.3.6
Installing cucumber-tag-expressions 2.0.4
Fetching cucumber-cucumber-expressions 10.3.0
Using diff-lcs 1.5.0
Using multi_test 0.1.2
Using ffi 1.16.3
Using rspec-support 3.12.1
Using bundler 2.4.10
Using docile 1.4.0
Using simplecov-html 0.12.3
Using simplecov_json_formatter 0.1.4
Using coderay 1.1.3
Using rexml 3.2.6
Using excon 0.104.0
Using faraday-em_http 1.0.0
Using faraday-em_synchrony 1.0.0
Using faraday-excon 1.1.0
Using faraday-httpclient 1.0.1
Fetching cucumber-html-formatter 6.0.3
Installing cucumber-gherkin 13.0.0
Fetching cucumber-core 7.0.0
Installing cucumber-html-formatter 6.0.3
Installing cucumber-core 7.0.0
Fetching cucumber-wire 3.0.0
Installing cucumber-wire 3.0.0
Fetching cucumber 4.0.0
Installing cucumber 4.0.0
Using aruba 0.14.14
Bundle complete! 22 Gemfile dependencies, 81 gems now installed.
Bundled gems are installed into `./vendor/bundle`
Error: Process completed with exit code 1.
```

</details>

Ruby all of them
Gem current

[...]

PR #1002, opened 2023-10-20
Consolidate Gemfiles; improve and simplify CI workflow

### LINKED ISSUE #1000: CI: Failures are prohibitively difficult to troubleshoot
The Cucumber and RSpec steps in CI redirect stderr to a temp file. This hides the output from the GitHub Actions console. As a result, when test execution fails (as recently happened in #995), it is impossible to see what is going wrong. Troubleshooting becomes a much more drawn out process.

E.g.

```
bundle exec rspec spec/ 2>> "${ALL_WARNINGS}"
```

The stderr capture is to facilitate the "check warnings" step, which fails the CI build if any tests output warnings. I assume this is to check for deprecations, so that these can be caught and corrected before impacting end users.

However, in the grand scheme, a deprecation warning is usually more of annoyance than a bug to end users. IMO avoiding these is not worth the trouble of complicating CI.

Perhaps a better way to guard against deprecations that are likely to eventually impact end-users is to add `ruby-head` to the build matrix. That way if a deprecation in a current version of Ruby is planned on becoming an error in a future version, it will fail the ruby-head build and maintainers will have a chance to correct it.

Alternatively, something like the [upload-artifact](https://github.com/actions/upload-artifact) action could be used to upload the captured stderr for troubleshooting purposes. This adds more complexity to the CI job, but if it works reliably, it could make troubleshooting possible while still maintaining the "check warnings" step.

mattbrictson:
The goal of this PR is to improve the long-term maintainability of vcr by greatly simplifying the CI infrastructure and improving its troubleshooting output.

### Improving troubleshooting output

Before, cucumber and rspec commands had their stderr output redirected to a temporary file. The contents of this file were never explicitly printed or archived. If the command failed (as recently happened in #995), maintainers would be flying blind in terms of understanding the failure.

With the changes of this PR, the redirection of stderr still exists, but if a command fails, the stderr output is explicitly printed to the console prior to the CI job exiting. This allows maintainers to see what went wrong in failure scenarios.

Fixes #1000.

### Simplifying Gemfiles and CI infrastructure

Before, the CI workflow was organized as a 1:1 conversion of a previous Travis CI setup, which was structured as one big CI script and multiple Gemfiles.

Over the years, the vcr project dropped the `appraisal` gem in favor of manually managing Gemfile variants. The project also at times tested JRuby and Rubinius, before removing them from CI. These changes and experiments have left a lot of cruft in the testing infrastructure.

This PR performs several cleanup operations:

1. Removes extra complexity of supporting JRuby and Rubinius. These runtimes were not being tested, so the complexity did not serve a purpose.
2. Removes separate Gemfiles. The `Gemfile.faraday-1.0.0` file, for example, had an overlapping version specification with the one in `vcr.gemspec`, effectively making it redundant. In the past, CI was configured to run against different major versions of some key dependencies, but that is no longer the case. Now the multiple Gemfiles do not have any benefit. This PR merges them all into a single Gemfile.
3. Splits the rspec, cucumber, and yard CI steps into parallel jobs. This is something that is easy to do in GitHub Actions that was probably prohibitively difficult in TravisCI. The parallel jobs increase performance and have other benefits.
4. Moves the stderr output redirection and warning detection logic out of the CI workflow and into a dedicated bash script.

My hope is that with fewer Gemfiles to maintain, and a more familiar GitHub Actions workflow, this PR will make vcr more approachable to maintainers and contributors going forward.


vcr.gemspec
```diff
@@ -20,25 +20,4 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.required_ruby_version = ">= 2.7"
-
-  spec.add_development_dependency "bundler", "~> 2.0"
```
mattbrictson, line 24: 🗒️ These dependencies are now declared in the Gemfile.

.github/workflows/test.yml
```diff
@@ -1,69 +1,56 @@
...
+      - name: rspec
+        run: ./script/fail_if_warnings rspec spec/
+  doc-coverage:
+    name: "Doc coverage"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: "ruby"
```
mattbrictson, line 53: 🗒️ I assumed that running the yard coverage check against a matrix of multiple versions of Ruby was overkill. I changed it to run on the latest version of Ruby only.

CONTRIBUTING.md
```diff
@@ -16,11 +16,3 @@ bundle exec rake
 VCR uses [RSpec](http://github.com/rspec/rspec) for unit tests.  The specs are written in a very "focused" style, where each spec is concerned only with exercising the object under test, using mocks as necessary.  You can run the specs using `rake spec`.
 
 [Cucumber](http://cukes.info/) is used for end-to-end full stack integration tests that also function as VCR's documentation.
-
-## Problems running bundle install?
-
-If you get an error while running `bundle install`, it may be one of the "extras" gems which are not required for development. Try installing it without these gems.
-
-```console
-bundle install --without extras
```
mattbrictson, line 25: 🗒️ There is no `extras` group in the Gemfile, so this documentation was misleading. I removed it to prevent confusion.

olleolleolle (approved):
Thank you, @mattbrictson! I agree with all the changes, and thanks also for all the context-preserving notes in the Description and comments of this PR.

PR #1003, opened 2023-11-06
Declare a gem dependency on the Ruby Base 64 gem

olleolleolle:
For compatibility with Ruby 3.4 and to silence compatibility warnings present in Ruby 3.3, declare a dependency on the `base64` gem.

In addition, in order to not fail in CI, this change was made:

bild

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,18 +7,24 @@ on:
     types: [opened, synchronize, reopened]
 
 jobs:
+  install_os_deps:
+    name: "Install OS dependencies"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install OS dependencies
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install --assume-yes libcurl4-openssl-dev
+
   cucumber:
     name: "Cucumber / ${{ matrix.ruby-version }}"
+    needs: [install_os_deps]
     runs-on: ubuntu-latest
     strategy:
       matrix:
         ruby-version: ["3.2", "3.1", "3.0", "2.7"]
     steps:
       - uses: actions/checkout@v4
-      - name: Install OS dependencies
-        run: |
-          sudo apt-get update -qq
-          sudo apt-get install --assume-yes libcurl4-openssl-dev
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: ${{ matrix.ruby-version }}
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -7,18 +7,24 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install OS dependencies
+        run: |
+          sudo apt-get update -qq
+          sudo apt-get install --assume-yes libcurl4-openssl-dev
+
   cucumber:
     name: "Cucumber / ${{ matrix.ruby-version }}"
+    needs: [install_os_deps]
```
mattbrictson, line 21: @olleolleolle I don't think `needs` will work for this, because AFAIK each job is executed on an independent container. So the packages that `install_os_deps` installs will have has no effect on subsequent jobs, which run in different containers.

I think copy-pasting the "Install OS dependencies" step into each job is the easiest solution. GitHub Actions doesn't have a lightweight way to share steps across jobs. The only way I know of is to create a [composite action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action), which requires an entirely separate YAML file.

olleolleolle: When I didn't have the step, the doc coverage step actually failed at install time, in bundle install due to curl not being there.

olleolleolle: Was that incidental?

mattbrictson: @olleolleolle I think the behavior is hard to determine because there is a bundler cache involved. Depending on the state of the cache, `bundle install` might get skipped, which leads to the job succeeding incidentally. The latest commit on the master branch is [failing](https://github.com/vcr/vcr/actions/runs/6768208454/job/18392207885) with "An error occurred while installing curb (1.0.5), and Bundler cannot continue," which to me is an indication that `install_os_deps` is not working as intended.

One possible solution would be to run the `ruby/setup` action inside the `install_os_deps` job. That would update the bundler cache once gems with native extensions have been compiled (e.g. curb). Then when the other jobs run, they will skip the `bundle install` step and use the cache that has just been warmed. That would be enough to fix the docs job, probably. But for tests, as soon as a test tries to e.g. use the curb gem, we would likely see an error since native libraries are missing.

Ultimately I think the most reliable approach is to explicitly install the apt packages in every job. To DRY it up a little, we could extract it to a shell script like `script/install-apt-deps.sh`, and then each job could just have:

```
- run: ./script/install-apt-deps.sh
```

AlexWayfer:
I'm looking forward to a new release with it.

PR #1010, opened 2024-01-26
CI: add Ruby 3.3 to build matrix

olleolleolle:
This PR extends the CI build matrix.

After another release, I'd like to cut out the EOL versions, and raise the minimum requirement.

PR #1011, opened 2024-01-26
Hold back rack at 2.x, add gems required for Ruby 3.3

olleolleolle:
When preparing  #1010, it became clear that some additions to the development Gemfile would be good.

Add "Rack < 3" also meant that the Rack modernization work could wait until we'd gotten a release out.


lucaskanashiro:
Are you still planning to support rack >= 3?

olleolleolle:
@lucaskanashiro Yes, it's absolutely a plan. Let's see if there's an Issue signposting this. If not, I'll add one.

PR #1014, opened 2024-02-05
Drop runtime dependency on base64

Earlopain:
#1003 added `base64` as a runtime dependency. It's easy enough to replace, the `base64` is for the most part a wrapper around the `pack`/`unpack` methods.

Also see:
* https://github.com/lostisland/faraday/pull/1541
* https://github.com/bblimke/webmock/pull/1046
* https://github.com/rubocop/rubocop/pull/12313
* https://github.com/rack/rack/pull/2110
* etc.

I have made it a dev dependency since once test file uses it in a bunch of places and replacing all that doesn't make the code much better.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ gemspec
 gem "ethon", "0.15.0"
 
 gem "aruba", "~> 0.14.14"
+gem "base64"
 gem "bigdecimal"
 gem "codeclimate-test-reporter", "~> 0.4"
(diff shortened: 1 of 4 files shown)


dentarg:
Why? The abstraction is there for a reason!

olleolleolle (approved):
Thanks, @Earlopain!

olleolleolle (approved):
There, one fewer dependencies. A tiny one.

Issue #1012, opened 2024-01-29
tests: Drop module HeaderDowncaser and use Hash#transform_keys instead


olleolleolle:
**Is your feature request related to a problem? Please describe.**

[Ruby 2.5 added `Hash#transform_keys`](https://www.bigbinary.com/blog/ruby-2-5-adds-hash-transform_keys-method)

**Describe the solution you'd like**

I'd like to omit the mixin in https://github.com/vcr/vcr/blob/20e6c2d3f6f6499725ec585053a0dd48f7d1b7f8/spec/support/http_library_adapters.rb#L1

**Describe alternatives you've considered**

Keeping the thing. Well, we don't support Ruby 2.4 any longer. No need.

notapatch:
Hi [olleolleolle](https://github.com/olleolleolle) and all.

I've put a PR for the issue which you can, of course, review at your convenience(s)!

PR #1015, opened 2024-02-21
Replace HeaderDowncaser with Hash#transform_keys

notapatch:
VCR dropping support for Ruby 2.4 allows a VCR specific module, HeaderDowncaser, to be replaced by a Ruby standard library call Hash#transform_keys.

I didn't find any unit tests specific to the module but the support module was used throughout the RSpec test suite.

Reference

Ruby 2.5 introduced Hash#transform_keys
https://docs.ruby-lang.org/en/3.2/NEWS/NEWS-2_5_0.html

Close: #1012

### DIFF
--- a/spec/support/http_library_adapters.rb
+++ b/spec/support/http_library_adapters.rb
@@ -1,18 +1,6 @@
-module HeaderDowncaser
-  def downcase_headers(headers)
-    {}.tap do |downcased|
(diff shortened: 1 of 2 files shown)


olleolleolle (approved):
Thanks, @notapatch 🙏 Good one!

PR #1019, opened 2024-07-04
Make it possible to turn VCR on and off in a new thread

aglundahl:
#764 fixed so that VCR's `:turned_off` status for the main thread can be read in a new thread, even if the status changes after the new thread was created.

However, the solution also introduced the problem that the `:turned_off` status isn't used for other threads than the main thread. This effectively makes it impossible to turn VCR on and off in a new thread.

This change, instead, creates new contexts with `:turned_off` and `:ignore_cassettes` set to `nil` by default. When the values are read, they are only used if they explicitly have been set on the thread's context. Otherwise the main thread's context is used.

### DIFF
--- a/lib/vcr.rb
+++ b/lib/vcr.rb
@@ -335,10 +335,11 @@ module VCR
   # @see #turn_off!
   # @see #turned_off
   def turned_on?
-    linked_context = current_context[:linked_context]
-    return !linked_context[:turned_off] if linked_context
-
-    !context_value(:turned_off)
+    if !context_value(:turned_off).nil?
+      !context_value(:turned_off)
+    else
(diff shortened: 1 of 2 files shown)


olleolleolle (commented):
Well, this looks handy. Does it pass tests?

aglundahl:
> Well, this looks handy. Does it pass tests?

Just updated my branch. Yes, the tests still pass.

nvanoorschot:
I think this fix would solve an issue we encounter when using VCR in combination with parallel_tests. The `turned_off` state can leak from test into another.

olleolleolle:
@aglundahl This looks fantastic. Thank you for building it.

If you don't mind, could you rebase on top of latest?

PR #1024, opened 2024-07-21
Make VCR CI compatible with Faraday 1 and 2

### LINKED ISSUE #1001: CI: Webmock + Faraday 2.x is not being tested
I believe that Faraday 2.x + Webmock + VCR is a common setup, given that Faraday is the [most popular](https://www.ruby-toolbox.com/categories/http_clients) Ruby HTTP library, and using VCR with Faraday 2.x [requires the webmock gem](https://github.com/vcr/vcr/issues/918).

Given the assumed popularity of this configuration, it makes me a little nervous that it is not being tested in CI. The VCR gemspec is currently locked at Faraday 1.x.

Fully supporting Faraday 2.x + VCR (i.e. without Webmock) would be nice (#918), but if that is blocked or prohibitively difficult, it would be good to at least have test coverage of Faraday 2.x + Webmock + VCR in the meantime.

andrehjr:
Hi, 

I took a stab into making specs run against faraday 1 and 2. In the past, this repo was using https://github.com/thoughtbot/appraisal for stuff like this, It's also an option.

But If we want only to test against faraday 2, I can rework this PR and remove the faraday 1.x stuff. 

I was also thinking about having the same strategy to support newer rack versions, while keeping compatibility 🤔  What do you think? 

This closes #1001

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,11 +8,14 @@ on:
 
 jobs:
   cucumber:
-    name: "Cucumber / ${{ matrix.ruby-version }}"
+    name: "Cucumber / Ruby ${{ matrix.ruby-version }} / Faraday ${{ matrix.faraday }}"
     runs-on: ubuntu-latest
     strategy:
       matrix:
         ruby-version: ["3.2", "3.1", "3.0", "2.7"]
+        faraday: ["1.0", "2.0"]
+    env:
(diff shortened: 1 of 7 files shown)


olleolleolle:
I think this is a great start, and I thank you for it! I think it looks good and I will merge it.

olleolleolle (approved):
@andrehjr Thanks!

PR #1025, opened 2024-07-21
Add Rack 2 and Rack 3 to the CI

### LINKED ISSUE #1020: Support Rack 3
In order to follow Rack's evolution, this repo's test suite needs to run both rack 2 and 3.

I'd like for VCR to know and support Rack 2 and Rack 3.

**alternatives considered**

None, really.

See #1011 for where Rack was held back.

andrehjr:
Same approach as #1024, but now adding Rack 2 and Rack 3 to Github Actions. Huge combination of executions now, but I think it's better to run against those combinations while most projects are still upgrading between faraday/rack.

I also got some commits from #1010 to add Ruby 3.3 too 🙌  hope that's ok. We may also add ruby-head later too? 

Closes #1020

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -8,14 +8,16 @@ on:
 
 jobs:
   cucumber:
-    name: "Cucumber / Ruby ${{ matrix.ruby-version }} / Faraday ${{ matrix.faraday }}"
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
Grand!

PR #1033, opened 2024-08-20
Move faraday-multipart to test only

colinbruce:


olleolleolle:
Now released as v6.3.1.

PR #1041, opened 2024-11-19
Add funding_uri to gemspec

andrew:
I noticed that the project already has a FUNDING.yml. This PR adds the `funding_uri` to your gemspec to help increase visibility using the `bundle fund` command in Bundler.


olleolleolle (approved):
Thanks for thinking of this, @andrew!

PR #1043, opened 2025-01-11
Remove unnecessary "".respond_to?(:encoding) checks

andrehjr:
These checks were added back when String didn't have encoding. 

As VCR supports Ruby 2.7+ now, I don't think they are necessary anymore.

I've also pinned excon to < 1.0.0, It seems some webmock specs are failing with excon >= 1.0.0 which was recently released.


olleolleolle (approved):
Thanks, @andrehjr!

Good context given in the PR description, Ruby 2.7 is an old memory, now.

PR #1054, opened 2025-06-02
Allow writing custom metadata to cassettes

balvig:
## What

This allows saving and retrieving extra metadata using `VCR::Cassette#metadata`.

Its purpose is similar to that of `VCR::Cassette#originally_recorded_at` in storing and setting up state from the time of recording when replaying cassettes.

## Why

The specific use case I had was related to writing tests against an API that required a unique email for every test run:

```rb
def setup
  VCR.insert_cassette("example")
end

def teardown
  VCR.eject_cassette
end

def create_user
  make_http_request("/users", email: "user@example.com")
end

test "First time works" do
  create_user # Works first time
end

test "Second time doesn't" do
  create_user #  => Error: Email has already been taken
end
```

I solved this by generating a unique postfix every time a test was run, something like this:

```rb
def setup
  @user_email_postfix = SecureRandom.uuid
  cassette = VCR.insert_cassette("example")
  if cassette.recording?
    cassette.metadata["email_postfix"] = @email_postfix # store generated postfix when recording
  else
    @email_postfix = cassette.metadata["email_postfix"] # set postfix from metadata when replaying
  end
end

def teardown
  VCR.eject_cassette
end

def create_user
  make_http_request("/users", email: "user+#{@email_postfix}@example.com")
end

test "First time works" do
  create_user # => Created user+c0d6136e-7b22-4240-b5a6-8787786a3387@example.com
end

test "Second time also works" do
  create_user # => Created user+71d919d4-05b0-4b2f-9b64-2d371844dd9f@example.com
end
```

To replay these tests with the original postfix, I needed a way to store the postfix that was in effect at the time of recording.

## Anything else

Please let me know if I'm overlooking a more effective way to achieve this!
And thank you for creating VCR, it's been a massive help in my work. 🙂


features/cassettes/custom_metadata.feature
```diff
@@ -0,0 +1,80 @@
+Feature: Custom Metadata
+
+  Custom metadata can be used to store extra data alongside
+  the recordings for later retrievel using `VCR::Cassette#metadata`.
```
olleolleolle, line 4:
```suggestion
  the recordings for later retrieval using `VCR::Cassette#metadata`.
```

lib/vcr/cassette.rb
```diff
@@ -149,8 +149,13 @@ def recording?
...
         "http_interactions" => interactions_to_record.map(&:to_hash),
-        "recorded_with"     => "VCR #{VCR.version}"
-      }
+        "recorded_with"     => "VCR #{VCR.version}",
+        "metadata"          => metadata
+      }.reject { |k, v| v.empty? }
+    end
+
+    def metadata
+      @metadata ||= recording? ? {} : deserialized_hash['metadata']
```
olleolleolle, line 158: Perhaps we require something like this?

```suggestion
      @metadata ||= recording? ? {} : (deserialized_hash['metadata'] || {})
```

There already existed tests which had a `nil` value in metadata, so we had coverage of that branch of things.

olleolleolle (commented):
An interesting proposal!

balvig:
Thank you for the suggestions @olleolleolle! 👍

olleolleolle (approved):
Thanks for adding this capability, @balvig!

PR #1059, opened 2025-08-18
Replace `CGI.parse` with `URI.decode_www_form` for Ruby 3.5+ compatibility

### LINKED ISSUE #1057: CGI Dependancy in Ruby 3.5
CGI dependancy. In one of my gems I test against Ruby's head version as well and noticed VCR fails on head currently due to a change in CGI support. Exceptions like this get thrown:

```
/home/runner/work/lunchmoney/lunchmoney/vendor/bundle/ruby/3.5.0+2/gems/vcr-6.3.1/lib/vcr/configuration.rb:505:in 'Kernel#method': undefined method 'parse' for class '#<Class:CGI>' (NameError)

      self.query_parser = CGI.method(:parse)
```

I see this is accompanied by an error:
```
warning: CGI library is removed from Ruby 3.5. Please use cgi/escape instead for CGI.escape and CGI.unescape features.
```

I think the fix for this is just adding cgi as an explicit gem dependancy instead of requiring the standard library include it.

Ruby 3.5
Gem 6.3.1

mmenanno:
## Summary
- Adds `cgi` as an explicit runtime dependency to fix compatibility with Ruby 3.5+
- Resolves issue where VCR fails due to CGI being removed from Ruby standard library

## Problem
Ruby 3.5 removed CGI from the standard library, causing VCR to fail with:
`undefined method 'parse' for class '#Class:CGI'`

This occurs because VCR uses `CGI.method(:parse)` in `lib/vcr/configuration.rb:505` for query string parsing.

## Solution
Added `spec.add_runtime_dependency 'cgi'` to `vcr.gemspec` to ensure the CGI library is available across all supported Ruby versions.

## Test plan
- [x] Verified gemspec builds successfully with `gem build vcr.gemspec`
- [x] Confirmed the change addresses the specific error mentioned in the issue

Fixes #1057

### DIFF
--- a/lib/vcr/configuration.rb
+++ b/lib/vcr/configuration.rb
@@ -1,6 +1,5 @@
 require 'vcr/util/hooks'
 require 'uri'
-require 'cgi'
 
 module VCR
   # Stores the VCR configuration.
@@ -142,7 +141,7 @@ module VCR
     # The `#==` method must return true if both objects represent the
     # same query string.
     #
-    # This defaults to `CGI.parse` from the ruby standard library.
+    # This defaults to `URI.decode_www_form` from the ruby standard library.
     #
     # @overload query_parser
     #  @return [#call] the current query string parser object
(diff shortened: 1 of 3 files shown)


olleolleolle (approved):
@mmenanno Thanks!

PR #1071, opened 2026-05-06
Use new style of hash notation

Yegorov:
Closes #1068

### DIFF
--- a/docs/cassettes/format.md
+++ b/docs/cassettes/format.md
@@ -81,7 +81,7 @@ VCR.use_cassette('example') do
 end
 ```
 
-_And_ I am configuring with "\<configuration\>" using "\<http_lib\>"
+_And_ I am configuring with "&lt;configuration&gt;" using "&lt;http_lib&gt;"
 
 _When_ I successfully run `ruby cassette_yaml.rb 'Hello'`
 
--- a/docs/cassettes/format.md
+++ b/docs/cassettes/format.md
@@ -81,7 +81,7 @@ VCR.use_cassette('example') do
 end
 ```
 
-_And_ I am configuring with "<configuration>" using "<http_lib>"
+_And_ I am configuring with "\<configuration\>" using "\<http_lib\>"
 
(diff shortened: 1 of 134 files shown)


Yegorov:
@olleolleolle can you take a look, please?

docs/cassettes/format.md
```diff
@@ -81,7 +81,7 @@ VCR.use_cassette('example') do
 end
 ```
 
-_And_ I am configuring with "<configuration>" using "<http_lib>"
+_And_ I am configuring with "\<configuration\>" using "\<http_lib\>"
```
olleolleolle: The special markers in Cucumber, perhaps they must remain as they're written? Or don't they?

Yegorov: @olleolleolle Yes, but this is necessary for correct rendering in Markdown.

And by the way, for some reason it doesn't work (https://yegorov.github.io/vcr/#/cassettes/format?id=requestresponse-data-is-saved-to-disk-as-yaml-by-default):


Alternatively, we can replace this in the documentation manually (but it will revert back to the next time you rebuild doc).
This conversion occurs in the script `./script/doc` [gherkin2markdown](https://github.com/vcr/vcr/blob/6f376c11c23eee7fd029f7735ac40ade2855fcd0/script/doc#L6) and, accordingly, the character replacement [`<,>`](https://github.com/raviqqe/gherkin2markdown/blob/76698d1d079970f301aa905b76979471ab618aa1/renderer/renderer.go#L130-L131).

~It would probably be a good idea to make a PR to replace `<` with `&lt;` and `>` with `&gt;` for [gherkin2markdown](https://github.com/raviqqe/gherkin2markdown) tool.~
Maybe we should try to fix this in docsify, `\<` is also the correct escape syntax in Markdown.

olleolleolle (commented):
The diff looks alright, perhaps there are formatting details in the Cucumber files which needs to stay the same (the `<foo>` syntax). Let's now run the test suite.

olleolleolle (approved):
Let's use this!

PR #1073, opened 2026-05-13
Fix frozen string literal warning when recompressing response

cgunther:
Without this, the following warning is output:
```
lib/vcr/structs.rb:404: warning: literal string will be frozen in the future (run with --debug-frozen-string-literal for more information)
```

Follow up to #832.


olleolleolle (approved):
Golden, thanks!

PR #1075, opened 2026-05-28
Support ruby 3.4

fkmy:
Add Ruby 3.4 to the CI test matrix and fix the issues that surfaced.

- Add `ostruct` to the Gemfile — Ruby 3.4 emits a warning that it will no longer be a default gem from Ruby 4.0.0:
  warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 4.0.0.
- Silence `literal string will be frozen` warnings around `force_encoding` and `gsub!` (chilled literals on Ruby 3.4+).
- Format cassette options with the legacy `{:key=>val}` style so debug log output stays consistent (Ruby 3.4 changed `Hash#inspect` for symbol keys).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        ruby-version: ["3.3", "3.2", "3.1", "3.0", "2.7"]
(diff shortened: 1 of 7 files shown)


olleolleolle (approved):
Thanks, @fkmy!

PR #1076, opened 2026-05-29
Add Ruby 4.0 to CI test matrix

fkmy:
The CGI library is no longer bundled in Ruby 4.0. `CGI.escape` is now available via `cgi/escape`.

ref: https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/#stdlib-compatibility-issues


olleolleolle (approved):
All green, thanks!

PR #1077, opened 2026-05-31
Refresh Ruby support matrix in README

fkmy:
- Add MRI 3.4 and 4.0 to the list of tested versions
- Drop MRI 2.6 from the list of tested versions


olleolleolle (approved):
Thanks!

